PR #9, opened 2015-05-23
replace sdk/widget (removed in firefox 38) with sdk/ui/button/toggle.

gamen:
I wasn't too happy to see my newly discovered extension break only days later, so I fixed it. Though I didn't touch showWidgetPanel; it's not used, so why widen the scope of the commit?


bwinton:
Awesome!  Thank you!  I'll do my best to review this later today…

lib/main.js
```diff
@@ -20,34 +20,17 @@ var notifications = require('sdk/notifications');
 var {setTimeout, clearTimeout} = require('sdk/timers');
 var snooze = require('./snooze');
 var tabs = require('sdk/tabs');
-var {Widget} = require('sdk/widget');
+var {ToggleButton} = require("sdk/ui/button/toggle");
```
bwinton: Style nit: single quotes, please.  :)

gamen: nit granted

lib/main.js
```diff
@@ -20,34 +20,17 @@ var notifications = require('sdk/notifications');
...
-         '          background-image: url("' + data.url(image.replace(".png", "@2x.png")) + '");\n' +
-         '        }\n' +
-         '      }\n' +
-         '    </style>\n' +
-         '  </head>\n' +
-         '  <body></body>\n' +
-         '</html>\n';
+  return {
+    "16": data.url(image),
+    "32": data.url(image.replace(".png", "@2x.png"))
```
bwinton: So, this doesn't look quite right on Retina Mac…
There are a whole new set of icons in https://github.com/bwinton/SnoozeTabs/commit/bdc230b3a9cf3593e99508924fdd7fded158395a (which I _just_ pushed)  Could you try using those, and I'll see if it works better on my end?

bwinton:
(Other than that, it looks great!  Thank you for the work!  :smiley:)

lib/main.js
gamen: The ActionButton doesn't seem to work; the panel doesn't pop up. Switching back to a ToggleButton works though. See my toggle_button branch... https://github.com/gamen/SnoozeTabs/tree/toggle_button

bwinton: Yeah, you're totally right.  We should definitely use a ToggleButton.  I was thinking that you could just grab the new icons from the commit I posted, and add those to your master branch (and return them from this function).  :)

gamen: And I've done so... except I just eliminated the makeContent function entirely as you did.

PR #25, opened 2016-11-22
Switch to jpm from cfx for add-on build

lmorchard:
Starting to explore this add-on a bit, and figured I'd try switching it over to jpm as a first task. You *should* be able to get this all going with:

```
npm install
grunt build
```


pdehaan:
Should we switch to [**jpm**](http://npm.im/jpm), or use [**web-ext**](http://npm.im/web-ext)?

bwinton:
We should use [web-ext](https://github.com/bwinton/SnoozeTabs/tree/webext).  😉

lmorchard:
Yeah, I just kind of stumbled in the door and hadn't seen the web-ext branch yet

PR #30, opened 2016-12-05
React refactor and management UI

lmorchard:
Still working on this, but figured I'd toss it on the PR queue for a peek so the build process changes aren't as much a surprise. Seems way simpler than I expected, though


lmorchard:
I've gone slightly mad and reworked everything to use React. But, the date/time picker seems to work now, and things seem cleaner IMO. Moving on to the management panel next

lmorchard:
One more bit of madness - starting to poke at the management UX, but got tired of always having to navigate the webextension button. Added a [react-storybook](https://github.com/storybooks/react-storybook) setup that lets me rig up the UI components in a hot-loader to edit & tweak more quickly

src/popup/snooze-content.js
```diff
@@ -0,0 +1,115 @@
...
+function fetchEntries() {
+  browser.storage.local.get().then(items => {
+    setState({ entries: Object.values(items || {}) });
+  });
+}
+
+function init() {
+  browser.storage.onChanged.addListener((changes, area) => {
+    // TODO: granularly apply the changes, rather than triggering a refresh?
+    if (area === 'local') { fetchEntries(); }
```
bwinton: Should this also render, or is that handled somewhere else?

src/popup/snooze.scss
```diff
@@ -0,0 +1,274 @@
...
+  max-height: 190px;
+}
+
+.rc-calendar-footer-btn {
+  display: flex;
+  flex-direction: row;
+}
+.rc-calendar-footer-btn > a {
+  flex-grow: 1;
+}
```
bwinton: This file is starting to get kinda long.  We should refactor it sometime.

bwinton: (Heh.  I couldn't find my previous comment.  Ignore this one.)

rollup.config.js
```diff
@@ -0,0 +1,29 @@
...
+      babelrc: false,
+      exclude: 'node_modules/**',
+      presets: [ [ 'es2015', { modules: false } ], 'stage-0', 'react' ],
+      plugins: [ 'external-helpers' ]
+    }),
+    cjs({
+      exclude: 'node_modules/process-es6/**',
+    }),
+    globals(),
+    replace({ 'process.env.NODE_ENV': '"development"' }),
```
lmorchard: Kind of an in-case-we-need-it thing that I copypasta'd from the rollup example I found while setting this all up. We could enable debugging stuff & things like a super-short snooze time for a dev build - and then turn that stuff off in an automated prod build.

PR #38, opened 2016-12-13
fix: Get the build script working on Windows.

### LINKED ISSUE #31: Ensure build process works on Windows
In #30, I've introduced a new build process based on npm scripts. There are some issues in just running `npm start` on Windows. Need to circle back and make sure there's a good solution.

bwinton:
Expand out `src/{manifest.json,icons}` to `src/manifest.json src/icons`.
Add a `-p` flag to `onchange` to poll (because inotify doesnʼt work).
Add a new target that doesnʼt call `run` or `storybook` (since they don't work yet).

Fixes #31.


README.md
```diff
@@ -8,6 +8,10 @@ An add-on to let you snooze your tabs for a while.
   * Starts a file watcher to rebuild on changes
   * Launches Firefox Dev Edition with the extension, reloaded on changes
 
+* If youʼre on Windows, youʼll need to use `npm start-win`
```
lmorchard: This should be `npm run start-win`, because it's not one of the "built-in" npm script short cuts.

lmorchard:
Taking a look at this on my Windows machine. Seeing some errors, but I'm trying it on the GitHub for Windows bash shell. Going to try getting the Linux subsystem running and see if I get a different result

lmorchard:
As a follow-on to this, I might take a look at getting this stuff working also under the bash shell installed by GitHub for Windows - I think it's a mingw thing? Seems like [npm tries to run scripts under CMD on Windows](https://github.com/npm/npm/issues/9420) even though I'm in a bash shell, which results in all kinds of shenanigans. Also not sure why the storybook server seems broken :/

bwinton:
The storybook server is broken because of [this bug](https://wpdev.uservoice.com/forums/266908-command-prompt-console-bash-on-ubuntu-on-windo/suggestions/16109719-enable-network-connection-enumeration) in the Windows subsystem.  It sounds like it's fixed in a more recent insider build, so I'll give it a try when I get that update.  🙂

PR #47, opened 2016-12-14
refactor: Use a single wake-up alarm for all tabs

### LINKED ISSUE #33: Simplify use of timers
Should be able to have the extension use a single timer - just the next time it needs to wake up and de-snooze a tab. Also need to ensure that timer is set on startup.

lmorchard:
- Switch to a single alarm to wake up and check for (over)due tabs,
  rather than setting an alarm for each tab

- Set wake-up alarm on init

- More logging chatter when NODE_ENV === 'development'

- Misc reorg & tidying to make init flow clearer

Fixes #33

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -9,62 +9,81 @@
 import moment from 'moment';
 import { times, timeForId } from './lib/times';
 
-browser.runtime.onMessage.addListener(handleMessage);
+const DEBUG = (process.env.NODE_ENV === 'development');
+const WAKE_ALARM_NAME = 'snooze-wake-alarm';
+
+function log(...args) {
+  if (DEBUG) { console.log('SnoozeTabs (BE):', ...args); }  // eslint-disable-line no-console
+}
+
+function init() {
+  log('init()');
+  browser.alarms.onAlarm.addListener(handleWake);
+  browser.notifications.onClicked.addListener(handleNotificationClick);
+  browser.runtime.onMessage.addListener(handleMessage);
+  handleWake().then(updateWakeAlarm);
+}
 
 function handleMessage({op, message}) {
-  console.log('backgroundMessage', op, message);  // eslint-disable-line no-console
-  switch(op) {
-    case 'schedule':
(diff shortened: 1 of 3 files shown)


src/background.js
```diff
@@ -9,60 +9,81 @@
...
+    browser.storage.local.remove(idForItem(message)).then(updateWakeAlarm),
+  update: message =>
+    messageOps.cancel(message.old).then(() => messageOps.schedule(message.updated))
+};
+
+function updateWakeAlarm() {
+  return browser.alarms.clearAll()
+    .then(() => browser.storage.local.get())
+    .then(items => {
+      const times = Object.values(items).map(item => item.time);
```
bwinton: Isn't that the same as `Object.keys(items)` (because we're storing the items with their time as the key)?
(Or `items.map(k => +k)` at worst…)

src/popup/snooze-content.js
```diff
@@ -66,6 +72,7 @@ function openSnoozedTab(item) {
     active: true,
     url: item.url
   });
+  cancelSnoozedTab(item);
```
bwinton: I do wonder a little about this…  Was that part of the spec?  If not, let's run it by @sevaan to see what he thinks.

lmorchard: Hmm, that's rather unrelated now that I think about it. I think I slipped it in as something I thought was missing - i.e. a snoozed tab waking up should probably get removed from the list. But, I can pull that out and handle it in another PR.

src/background.js
lmorchard: Well, the keys are strings and *could* conceivably be something other than the time at some point. Figured it was better to err on the side of being explicit about what I was pulling out of the data in storage

lmorchard:
Updated the commit to a) remove the cancel on click, b) rebase against the branch, and c) get everything passing the new lint roles

src/background.js
bwinton: 👍

Issue #50, opened 2016-12-15
Ensure date selection disallows values in the past


lmorchard:
You can currently pick any date/time in the past. That doesn't make much sense.

bwinton:
Pinging @sevaan for ideas on what to do here…  🙂

lmorchard:
> Pinging @sevaan for ideas on what to do here… 🙂

FWIW, I think he may be out this week.

sevaan:
I'm back at it this week!

Can we just [disable](https://firefoxux.github.io/StyleGuide/#/buttons) the Snooze button anytime the date/time is inactive?

bwinton:
https://twitter.com/bwinton/status/791337794874511360 seems relevant.  😉

(Although yes, I believe that wouldnʼt be _toooo_ hard in this case…)

Issue #36, opened 2016-12-12
Add `Calendar` panel.


bwinton:
So that we can use the same code to choose times for new items as well as existing items.

PR #51, opened 2016-12-16
Constrain date/time selection to future

lmorchard:
This one's a twofer - #36 for wrapping the Calendar control and then #50 to ensure that calendar will only allow values in the future.

### DIFF
--- a/src/lib/components/MainPanel.js
+++ b/src/lib/components/MainPanel.js
@@ -4,8 +4,7 @@ import classnames from 'classnames';
 import moment from 'moment';
 import { times, timeForId } from '../times';
 
-import Calendar from 'rc-calendar';
-import TimePickerPanel from 'rc-time-picker/lib/Panel';
+import DatePickerPanel from './DatePickerPanel';
 
 export default class MainPanel extends React.Component {
   constructor(props) {
@@ -19,7 +18,6 @@ export default class MainPanel extends React.Component {
   render() {
     const { id, active, switchPanel } = this.props;
     const { datepickerActive } = this.state;
-    const timePickerElement = <TimePickerPanel />;
 
     return (
       <div>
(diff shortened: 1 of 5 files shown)


src/lib/components/MainPanel.js
```diff
@@ -66,18 +61,12 @@ export default class MainPanel extends React.Component {
...
   closeTimeSelect() {
     this.setState({ datepickerActive: false });
   }
 
-  confirmTimeSelect() {
-    const { dateChoice } = this.state;
+  confirmTimeSelect(dateChoice) {
     const { scheduleSnoozedTab } = this.props;
-
     if (!dateChoice) { return; }
```
bwinton: So, I was able to set a time in the past by going to tomorrow, picking 9am, then going back to today.  I think we should probably check for that somewhere before here, but checking that the date/time is valid in `confirmTimeSelect` also seems like a reasonable thing to do.  😉

bwinton: (It also looks like setting an alarm for a time in the past doesn't ever fire the trigger, so we should _really_ check for that before setting the alarm.)

src/popup/snooze.scss
```diff
@@ -281,3 +282,7 @@ li.option {
 .rc-calendar-footer-btn > a {
   flex-grow: 1;
 }
+
+.rc-time-picker-panel-select ul li {
+  justify-content: center
```
lmorchard: Fixed, and filed #53 for auto-linting on commits

src/lib/components/MainPanel.js
lmorchard: > So, I was able to set a time in the past by going to tomorrow, picking 9am, then going back to today.

Hmm, I'll have to think how to finesse that exactly. Display a warning? Force the time to now? Just not accept a click on the snooze button until the time is valid?

> (It also looks like setting an alarm for a time in the past doesn't ever fire the trigger, so we should really check for that before setting the alarm.)

Huh, that's weird. [This *should* be catching all tabs with times before the current time](https://github.com/bwinton/SnoozeTabs/blob/master/src/background.js#L65). Though, of course, if we don't have a wake time anytime soon (i.e. because it's in the past), that code will never fire. Maybe the wake timer also needs to be guarded so it never get set for any past time - e.g. `Math.max(snoozeTime, now)` or something

lmorchard: Just did a quick hacky update to try to ensure the wake timer is no earlier than 5 seconds from now.

lmorchard: Okay, one more update for this - I added a disabled state & style for the snooze button that triggers when the selected time is in the past. Just did a quick & dirty opacity drop, but we can refine that if it's not copacetic

PR #58, opened 2016-12-20
fix: Update the wake alarm after we fire a notification.

bwinton:
Just a quick fix, I'm fairly sure we want to re-schedule the next alarm when the current one fires, so I added a call to do that.  (Also I printed out the console message in a friendlier format for ease of debugging/inspecting. 😉)


lmorchard:
Oh, derp - I should have done this to begin with!

PR #65, opened 2016-12-30
feat: Handle snoozing tabs until next open.

### LINKED ISSUE #41: Snooze until next browser restart
Implement a snooze option that's not a time, but instead is triggered when the browser is quit and later restarted.

bwinton:
* Add shx to run shell commands on Windows.
* Add a 'next' time to open tabs on next open.
* Change the id to include the url so that we can save more than one tab for the same time.
* Add an id so that the package we build can be tested in Nightly/Developer Edition.

Fixes #41.


bwinton:
@sevaan: An icon for this new menu item would be a nice touch…  🙂

sevaan:
SVG for an icon can be found here: http://i.sevaan.com/190W2C1O0E0X

src/background.js
```diff
@@ -65,20 +76,30 @@ function handleWake() {
...
-        'iconUrl': browser.extension.getURL('link.png'),
-        'title': item.title,
-        'message': item.url
-      })).then(() => browser.storage.local.remove(idForItem(item)))
-    ));
+    return browser.windows.getAll({
+      windowTypes: ['normal']
+    }).then(windows => {
+      let windowIds = windows.map(window => window.id);
+      return Promise.all(due.map(([id, item]) => {
```
pdehaan: ESLint is whining because `id` is apparently unused, but since we're destructuring an array here, not sure if we should do an inline `// eslint-disable-lint no-unused-vars` or if we can do something classier:

```
/Users/pdehaan/dev/github/mozilla/SnoozeTabs/src/background.js
  85:36  warning  'id' is defined but never used  no-unused-vars
```

src/lib/components/ManagePanel.js
```diff
@@ -55,9 +55,22 @@ export default class ManagePanel extends React.Component {
...
+    }
+    return moment(time).format('ddd, MMM D') || 'Later'
+  }
+
+  getTime(time) {
+    if (time === 'next') {
+      return 'Firefox opens';
+    }
+    return moment(time).format('[@] ha') || '';
+  }
```
pdehaan: Sorry, I don't know where to pin this comment, but the sharp edge case here seems to be managing a "Next time Firefox opens" snoozed tab, when you get back into calendar view.

firefox_developer_edition_and_feat__handle_snoozing_tabs_until_next_open__by_bwinton_ _pull_request__65_ _bwinton_snoozetabs_and_mozilla_-_calendar_-_week_of_jan_9__2017

firefox_developer_edition_and_feat__handle_snoozing_tabs_until_next_open__by_bwinton_ _pull_request__65_ _bwinton_snoozetabs_and_mozilla_-_calendar_-_week_of_jan_9__2017

pdehaan: Similarly, if you've snoozed a tab to a date in the future and then edit the time, you seemingly can't edit the reminder for "Next time Firefox opens".

bwinton: Uh, file a followup bug for that (in the Initial Release milestone), I think.  We'll need a new design from Sevaan.

pdehaan: Filed as #85.

src/background.js
```diff
@@ -21,15 +21,26 @@ function init() {
   browser.alarms.onAlarm.addListener(handleWake);
   browser.notifications.onClicked.addListener(handleNotificationClick);
   browser.runtime.onMessage.addListener(handleMessage);
-  handleWake();
+  browser.storage.local.get().then(items => {
+    const due = Object.entries(items).filter(item => item[1].time === 'next');
```
pdehaan: Not sure if we should abstract the `'next'` constant into a separate file or something.
I think it's used in about 8 places in this PR (probably not that we're thinking of refactoring this ever, but still, thought I'd throw this out there).

src/background.js
```diff
@@ -65,20 +76,30 @@ function handleWake() {
...
-    ));
+    return browser.windows.getAll({
+      windowTypes: ['normal']
+    }).then(windows => {
+      let windowIds = windows.map(window => window.id);
+      return Promise.all(due.map(([, item]) => {
+        let createProps = {
+          active: false,
+          url: item.url,
+          windowId: windowIds.indexOf(item.windowId) === -1 ? undefined : item.windowId
```
pdehaan: FWIW (and it ain't "W" much), ES2015 has an [`Array.includes()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) method which returns a Boolean if the array includes the specified element:

```
windowId: windowIds.includes(item.windowId) ? item.windowId : undefined
```

src/background.js
```diff
@@ -21,15 +21,26 @@ function init() {
   browser.alarms.onAlarm.addListener(handleWake);
   browser.notifications.onClicked.addListener(handleNotificationClick);
   browser.runtime.onMessage.addListener(handleMessage);
-  handleWake();
+  browser.storage.local.get().then(items => {
```
pdehaan: Not sure if we want to [eventually] add `.catch()` blocks for all the promises too, in case of unexpected errors...

There is an ESLint plugin (naturally), appropriately named [**eslint-plugin-promise**](http://npm.im/eslint-plugin-promise) which makes this pretty easy:

```sh
$ eslint --max-warnings=0 src

/Users/pdehaan/dev/github/mozilla/SnoozeTabs/src/background.js
   24:3  error  Expected catch() or return  promise/catch-or-return
   32:5  error  Expected catch() or return  promise/catch-or-return
  129:5  error  Expected catch() or return  promise/catch-or-return

/Users/pdehaan/dev/github/mozilla/SnoozeTabs/src/popup/snooze-content.js
   32:3  error  Expected catch() or return  promise/catch-or-return
  103:3  error  Expected catch() or return  promise/catch-or-return

✖ 5 problems (5 errors, 0 warnings)
```

```js
  "plugins": [
    "promise",
    "react"
  ],
  "rules": {
    "promise/avoid-new": "warn",
    "promise/catch-or-return": "error",
    "promise/no-callback-in-promise": "warn",
    "promise/no-promise-in-callback": "warn",
    "promise/no-return-wrap": "error",
    "promise/param-names": "error",
    ...
```

src/background.js
bwinton: I thought about it, but didn't bother this time.  Probably worth doing, though.

src/background.js
bwinton: This feels like a big enough change already.  How about we do that in a followup pr?  :)

src/background.js
bwinton: Yeah, that seems better.  Fixed!

src/background.js
lmorchard: nit: may want to move this big block into its own function, if only to keep that `init()` function small & clean. But, not a review-stopper.

src/background.js
```diff
@@ -65,20 +76,30 @@ function handleWake() {
   const now = Date.now();
   log('woke at', now);
   return browser.storage.local.get().then(items => {
-    const due = Object.values(items).filter(item => item.time <= now);
+    const due = Object.entries(items).filter(entry => entry[1].time <= now);
```
lmorchard: nit: Seems like the keys from `items` never gets used here, so would `values` still work?

src/background.js
```diff
@@ -65,20 +76,30 @@ function handleWake() {
...
-        url: item.url,
-        windowId: item.windowId
-      }).then(tab => browser.notifications.create(`${item.windowId}:${tab.id}`, {
-        'type': 'basic',
-        'iconUrl': browser.extension.getURL('link.png'),
-        'title': item.title,
-        'message': item.url
-      })).then(() => browser.storage.local.remove(idForItem(item)))
-    ));
+    return browser.windows.getAll({
```
lmorchard: Thinking through this bit so I understand it: Looks like it reuses windows by ID when they exist, but open new ones otherwise? Not sure what it did before, so this seems like a nice bugfix.

src/icons/Next Open.svg
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
lmorchard: I see the icon, yay! :+1:

lmorchard:
Going to go ahead & merge this and we can clean other things up later if needed because they're more fussy nits than anything real

package.json
```diff
@@ -29,6 +29,7 @@
     "rollup-plugin-node-resolve": "^2.0.0",
     "rollup-plugin-replace": "^1.1.1",
     "sass-lint": "^1.10.2",
+    "shx": "^0.2.1",
```
bwinton: I'm running on Windows, so I can verify it works on at least one configuration.  😉

lmorchard: Cool, so with our computer powers combined, this seems like a success

src/background.js
bwinton: Before it threw an error "invalid windowId", and didn't open anything.  😉

src/background.js
bwinton: Yes…  I think I was using the keys for a while, but they should be the same as `entry.id` (or something).  I'll throw that in the next PR.

bwinton: Oh, wait!  I don't think we can, because we use `due.map(entry => entry[0])` down on line 100, so `due` needs to be the entries…

PR #78, opened 2017-01-07
Strip any leading www. from displayed domain names

### LINKED ISSUE #71: Strip any leading "www." from displayed domain names?
### Steps to reproduce:

1. Surf the net and snooze some tabs.
1. Manage your snoozed tabs via Snooze Tabs > Manage Snoozed Tabs.

### Actual results:
Several of my snoozed tabs have the "www." subdomain, which can take up "valuable" space.
Not sure if we want to special case this boring subdomain and do something clever like `return parser.host.replace(/^www\./, '');` in [./src/lib/components/ManagePanel.js:53](https://github.com/bwinton/SnoozeTabs/blob/b2ea0b116224006ebb5cf82ad1f30c809c5c4683/src/lib/components/ManagePanel.js#L53).

It's a bit of a stretch, since it'd only impact really long domain names where the TLD may be truncated.
As you can see in the screenshot below "www.washingtonpost.com" is fine with the "www.", but the other "www.top10bestwebsitehosti..." gets truncated. 

are_we_rebels_yet__and_data_url_-_yahoo_search_results_and_really_long_domain_name_-_yahoo_search_results

### With the `www.` removing patch applied:

menubar_and_firefox_developer_edition_start_page_and_1__npm_start__node__and_managepanel_js_ _snoozetabs

pdehaan:
Fixes #71

### DIFF
--- a/src/lib/components/ManagePanel.js
+++ b/src/lib/components/ManagePanel.js
@@ -50,7 +50,7 @@ export default class ManagePanel extends React.Component {
     var parser = document.createElement('a');
     parser.href = url;
     if (parser.protocol.startsWith('http')) {
-      return parser.host;
+      return parser.host.replace(/^www\./, '');
     }
     return url;
   }


src/lib/components/ManagePanel.js
```diff
@@ -50,7 +50,7 @@ export default class ManagePanel extends React.Component {
     var parser = document.createElement('a');
     parser.href = url;
     if (parser.protocol.startsWith('http')) {
-      return parser.host;
+      return parser.host.replace(/^www\./, '');
     }
     return url;
```
bwinton: I wonder if we should just return the host here, and let the title distinguish the content…  I mean, it's not like we'll see much of the path anyways.

bwinton: (Uh, I can't read code.  Never mind me.  :D )

PR #86, opened 2017-01-11
feat: Flash the favicon of newly-woken tabs.

### LINKED ISSUE #43: Button badge when a tab is restored
We can't really bounce the toolbar button or decorate the newly active tab. Beyond that, it would be good to put an indicator on the toolbar button when a tab is restored - maybe a count? maybe just a dot?

### LINKED ISSUE #61: Snooze Tabs icon should not be coloured
The Snooze Tabs icon is currently orange in the toolbar. It should not have a colour like the other icons in the browser toolbar, and only turn orange when a snoozed tab is returned (if we can do the icon bounce).

bwinton:
Yeah, it's kinda hacky, but it seems to work, and there aren't any other good APIs for it.  😉

I should probably make a different PR first, but this was next in my queue, so let's go with it for now.
@sevaan: Can you let me know how often and for how long to flash the icon?

* Also, make the default icon grey.
* Also also, add an eslint-plugin-promise to check our promises.

Fixes #43.
Fixes #61.


src/background.js
```diff
@@ -16,11 +16,22 @@ function log(...args) {
   if (DEBUG) { console.log('SnoozeTabs (BE):', ...args); }  // eslint-disable-line no-console
 }
 
+var iconData;
```
pdehaan: `let` is the new `var`.
Not sure if we want to add `no-var` and `prefer-const` to the ESLint config.

bwinton: At the top level, they're the same!  😁  But adding stuff to the eslint config sounds like a good idea.

sevaan:
After playing around with a quick prototype, let's try switching the favicon 10 times, and show each icon for 0.5 seconds each.

src/background.js
bwinton: And fixed.

bwinton:
I also made it conform to @sevaan's suggestion on the number and times to flash.

sevaan:
We should also reflash the icon every minute for 10 minutes or until the user interacts with the tab.

bwinton:
I think that might be a followup PR.

lmorchard:
A little derailed today with meetings and such, will take a run through this tonight or first thing in the morning

src/manifest.json
```diff
@@ -14,7 +14,15 @@
...
   "version": "1.0",
-  "permissions": ["alarms", "contextMenus", "notifications", "storage", "tabs"],
+
+  "permissions": [
+    "alarms",
+    "contextMenus",
+    "notifications",
+    "storage",
+    "tabs",
+    "<all_urls>"],
```
lmorchard: That really is a permission name, huh? Weird!

Issue #67, opened 2017-01-06
In DevEdition's dark theme, the SnoozeTabs icon should be white


sevaan:
The SnoozeTabs icon in the dark theme for Firefox's DevEdition browser should be white.

![image 2017-01-06 at 12 08 59 pm](https://cloud.githubusercontent.com/assets/473531/21726034/f3d3e916-d408-11e6-98dc-b7419bd10e24.png)

bwinton:
This is currently impossible to do.  I filed [bug 1329242](https://bugzilla.mozilla.org/show_bug.cgi?id=1329242) to add an API or something to WebExtensions, which should let us fix this in the future, but until that's fixed, I vote we close this one as "unfixable"…  😢

bwinton:
It looks like [bug 1329242](https://bugzilla.mozilla.org/show_bug.cgi?id=1329242) has landed, so we should probably look into this at some point…

lmorchard:
Yeah, would be good to not keep closing dupes :)

SoftVision-EmilPasca:
I've checked this on the latest Nightly(56.0a1-20170630030203), the SnoozeTabs icon looks better, however, in my opinion I think we should keep the same color as the other toolbar icons.
![snoozetabs_icon_better](https://user-images.githubusercontent.com/20083658/27736893-bf8b92fa-5dae-11e7-9f9d-179cd11cc5f2.png)

terrycloth:
We could also maybe have a single icon image if there were an opaque background, to ensure sufficient visual contrast.

SoftVision-PaulOiegas:
@lmorchard John just fixed yesterday the Test Pilot button icon to work as expected on all 3 Firefox themes. Maybe we can do the same with the Snooze Tabs one which is not completely adapting to the themes?
Also, bug [1329242 ](https://bugzilla.mozilla.org/show_bug.cgi?id=1329242) that was blocking this, seems to be fixed now.

umpirsky:
In `57.0b4 (64-bit)`:

![snooze-tabs](https://user-images.githubusercontent.com/208957/31052658-72598b10-a68b-11e7-9add-0d71f7f8b738.gif)

MikkCZ:
I am not sure about how DevEdition looks now, but in Nightly 58 with the dark theme the icon is almost invisible. Sometimes I almost forgot to use this instead of pinning the tab, which I was used before installing this experiment.

umpirsky:
[...]

PR #93, opened 2017-01-13
fix: Disable the snooze button for special urls.

### LINKED ISSUE #59: Attempting to snooze special urls causes chaos…
`about:debugging` for instance, can't be restored (nor can `about:newtab` or some others).
So, we snooze it, and it goes away, but we can't open it from the Manage menu, and when its alarm fires, it fails to restore (and is thus lost forever), and the error prevents us from rescheduling the next alarm, so the whole thing breaks!  😢

Maybe we should check our ability to restore a tab before we snooze it?  Or just whitelist http/https urls and only deal with those?

bwinton:
We can only restore certain urls so let's not let people snooze them!
But we don't want to disable the button entirely, so if they click it, just show them the manage panel.

Fixes #59.

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -18,11 +18,36 @@ function log(...args) {
 
 let iconData;
 
+function updateButtonForTab(tabId, changeInfo) {
+  if (!changeInfo.url) {
+    return;
+  }
+  const url = changeInfo.url;
+  if (url.startsWith('http:') || url.startsWith('https:') || url.startsWith('file:') ||
+      url.startsWith('ftp:') || url.startsWith('app:')) {
+    browser.browserAction.setIcon({path: 'icons/Bell Icon.svg', tabId: tabId});
+  } else {
+    browser.browserAction.setIcon({path: 'icons/Disabled Bell Icon.svg', tabId: tabId});
+  }
+}
+
 function init() {
   log('init()');
   browser.alarms.onAlarm.addListener(handleWake);
   browser.notifications.onClicked.addListener(handleNotificationClick);
   browser.runtime.onMessage.addListener(handleMessage);
+  browser.tabs.onUpdated.addListener(updateButtonForTab);
+  browser.tabs.onCreated.addListener(tab => {
+    updateButtonForTab(tab.id, {url: tab.url});
+  });
+  browser.tabs.query({}).then(tabs => {
+    for (const tab of tabs) {
+      updateButtonForTab(tab.id, {url: tab.url});
+    }
+  }).catch(reason => {
+    log('init tabs query rejected', reason);
+  });
+
   if (!iconData) {
     fetch(browser.extension.getURL('icons/Color Bell Icon.png')).then(response => {
       return response.arrayBuffer();
@@ -32,6 +57,7 @@ function init() {
       log('init getUrl rejected', reason);
     });
   }
+
   browser.storage.local.get().then(items => {
     const due = Object.entries(items).filter(item => item[1].time === NEXT_OPEN);
     const updated = {}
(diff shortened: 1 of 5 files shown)


src/icons/Bell Icon.svg
```diff
@@ -1,28 +1,28 @@
...
-		l0.2,0.3L9,5.6c-0.3,0.2-0.7,0.3-1,0.5C6.7,7,5.8,8.3,5.3,9.8c-0.2,0.7-0.4,1.4-0.4,2.2v1.4c0,2-0.8,3.9-2.2,5.3
-		c-0.3,0.3-0.3,0.7-0.2,1c0.1,0.3,0.4,0.6,0.8,0.6h4.9h0.3v0.3c0,1.9,1.5,3.4,3.4,3.4s3.4-1.5,3.4-3.4v-0.3h0.3h4.9
-		c0.4,0,0.7-0.3,0.8-0.6C21.6,19.5,21.6,19.1,21.3,18.7z M12,1.8c0.9,0,1.6,0.7,1.6,1.6S12.9,4.9,12,4.9s-1.6-0.7-1.6-1.6
-		C10.5,2.5,11.1,1.8,12,1.8z M13.1,21.7c-0.3,0.3-0.7,0.5-1.1,0.5c-0.9,0-1.6-0.7-1.6-1.6v-0.3h0.3h2.5h0.3l0,0.3
-		C13.6,21,13.4,21.4,13.1,21.7z M14.8,11c0,0.3-0.1,0.6-0.4,0.9l-2.9,3.2h2.6c0.4,0,0.7,0.3,0.7,0.7s-0.3,0.7-0.7,0.7h-4
-		c-0.5,0-0.9-0.3-0.9-0.7v0c0-0.3,0.1-0.6,0.4-0.9l2.9-3.2H10c-0.4,0-0.7-0.3-0.7-0.7c0-0.4,0.3-0.7,0.7-0.7h3.9
-		C14.4,10.3,14.8,10.6,14.8,11L14.8,11z"/>
-</g>
-</svg>
+<?xml version="1.0" encoding="utf-8"?>
```
lmorchard: Hmm, so in Dev Edition, the disabled icon is lighter and the enabled icon is darker. That made me think the states were switched. But, this sounds like a variation on Issue #67. 

Maybe we want to think about changing the icon shape itself rather than the color? (i.e. circle-with-crossbar over the bell? just a random thought)

src/background.js
```diff
@@ -18,11 +18,36 @@ function log(...args) {
...
 function init() {
   log('init()');
   browser.alarms.onAlarm.addListener(handleWake);
   browser.notifications.onClicked.addListener(handleNotificationClick);
   browser.runtime.onMessage.addListener(handleMessage);
+  browser.tabs.onUpdated.addListener(updateButtonForTab);
+  browser.tabs.onCreated.addListener(tab => {
+    updateButtonForTab(tab.id, {url: tab.url});
+  });
+  browser.tabs.query({}).then(tabs => {
```
lmorchard: Not 100% sure what this bit does. Seems like it runs through all tabs and flips the button to enabled / disabled for each in turn. Does it just end up on the current tab in the end?

lmorchard: Maybe we want this here instead?
```
browser.tabs.query({currentWindow: true, active: true})
```

bwinton: It does, and we want it the way it is because we need to set the states for the existing tabs, and we don't get a `tabs.onUpdated` even when we switch tabs.

(i.e. If we start with two tabs, "about:home" and "about:addons", load the extension, we want it to be disabled on both.)

src/icons/Bell Icon.svg
bwinton: Yeah, I'm not sure we can fix #67.  At least, not yet.

Issue #46, opened 2016-12-14
Set up metrics code


lmorchard:
Along with [a plan](https://github.com/bwinton/SnoozeTabs/issues/45), we need to implement the code to measure things. This will be a user of [the proposed metrics broker library from Test Pilot](https://github.com/mozilla/testpilot/issues/1735).

Issue #45, opened 2016-12-14
Metrics plan


lmorchard:
Need to come up with a firm initial metrics plan. Notes from [acceptance criteria](https://github.com/bwinton/SnoozeTabs/issues/24) as a starting point:

* It should measure each time the snooze panel is opened
* It should measure each time a user choses to snooze
* It should measure which snooze duration is chosen each time
* it should measure custom snooze durations
* It should measure the rate at which snooze tabs wake up
* it should measure if users focus previously snoozed tabs
* it should measure if users close a previously snoozed tab without refocusing
* it should measure if users re-snooze a tab
* it should measure snooze cancellations
* it should measure clicks on currently snoozed tabs
* <strike>it should measure clicks into bookmarks folder from the panel</strike>

lmorchard:
Additional note, here: This plan should result in the production of `docs/metrics.md` in the repo, describing the metrics in a public place. Some examples: [minvid](https://github.com/meandavejustice/min-vid/blob/master/docs/metrics.md); [Tab Center](https://github.com/bwinton/TabCenter/blob/master/docs/metrics.md).

johngruen:
@lmorchard i think final UX does not have any bookmarks folder backing snooze tabs, so that last item is moot

ckprice:
Needs: what are the questions we're asking, and how do we know it's successful. Assigning to Wil and John.

PR #95, opened 2017-01-13
Set up metrics code (#45 & #46)

lmorchard:
Throwing this out there to start getting some eyes on it. I tried cribbing `docs/metrics.md` and the ping format [from fellow webextension-style experiment blok](https://github.com/mozilla/blok/blob/master/docs/metrics.md), but it could use some review & further tweaking.

Will poke some other Test Pilot folks next week for opinons - off the top of my head, @johngruen & @6a68 & @clouserw might have some?


lmorchard:
Also, the tests here don't actually test `background.js` - but as I tried getting into that, it seemed like I'd have to restructure a bunch of it into a submodule that could get mock `browser` APIs injected via `init()` and I figured that's too ocean-boiling-esque for this PR.

docs/metrics.md
```diff
@@ -0,0 +1,113 @@
...
+
+TBD
+
+## Data Collection
+
+### Server Side
+There is currently no server side component to Snooze Tabs.
+
+### Client Side
+Snooze Tabs will use Test Pilot's Telemetry wrapper with no batching of data.
```
wresuolc: We'll modify this to say something about @_6a68 's broker once we have more details on it

docs/metrics.md
```diff
@@ -0,0 +1,113 @@
...
+* A previously snoozed tab is closed without having been focused
+```js
+{
+  "event": "closed-unfocused",
+  "snooze_time": 1484345836165,
+  "snooze_time_type": "tomorrow"
+}
+```
+
+* A previously snoozed tab is snoozed again after waking
```
wresuolc: This implies that once a tab is snoozed you're forever tracking that fact about that tab?

lmorchard: Hmm, yeah, maybe we should be more specific that it's only tracked the first time it's resnoozed. Or, until the browser restarts - which is what the code actually does right now, because I don't persist the tracking anywhere permanent.

docs/metrics.md
```diff
@@ -0,0 +1,113 @@
...
+* A snoozed tab is clicked from the management panel
+```js
+{
+  "event": "clicked",
+  "snooze_time": 1484345836165,
+  "snooze_time_type": "tomorrow"
+}
+```
+
+A Redshift schema for the payload:
```
lmorchard: Also, this schema is a total guess, since I've not done one before and haven't seen where it plugs in

src/background.js
```diff
@@ -83,14 +85,23 @@ const idForItem = item => `${item.time}-${item.url}`;
...
+  cancel: message => {
+    Metrics.cancelSnoozedTab(message);
+    return browser.storage.local.remove(idForItem(message)).then(updateWakeAlarm);
+  },
+  click: message => {
+    Metrics.clickSnoozedTab(message);
+  },
   update: message =>
-    messageOps.cancel(message.old).then(() => messageOps.schedule(message.updated))
+    messageOps.cancel(message.old).then(() => messageOps.schedule(message.updated)),
```
bwinton: Should this get a separate Metrics ping, or did we want to force people to back-calculate which cancel/schedule pairs were updates and which weren't?

src/lib/components/MainPanel.js
```diff
@@ -68,6 +68,6 @@ export default class MainPanel extends React.Component {
   confirmTimeSelect(dateChoice) {
     const { scheduleSnoozedTab } = this.props;
     if (!dateChoice) { return; }
-    scheduleSnoozedTab(dateChoice);
+    scheduleSnoozedTab(dateChoice, 'pick');
```
bwinton: I think we should use the constant from `lib/times.js` here instead.  (And add one for `'week'` down below.)

src/background.js
```diff
@@ -33,6 +34,7 @@ function updateButtonForTab(tabId, changeInfo) {
 
 function init() {
   log('init()');
+  Metrics.init(window.BroadcastChannel, browser.tabs);
```
bwinton: So, I'm not really sure why we're passing in the BroadcastChannel here.  I suspect it's the right thing to do, but could you elaborate a little more on it (maybe in a comment in `metrics.js`)?

src/lib/metrics.js
```diff
@@ -0,0 +1,99 @@
+// Track recent tabs woken to see whether they're later closed or focused
+let unseenWakeHistory = {};
+let seenWakeHistory = {};
+
+// BroadcastChannel for sending metrics pings to Test Pilot add-on
+let pingChannel = null;
+
+// Name of the metrics BroadcastChannel expected by the Test Pilot add-on
+const TESTPILOT_TELEMETRY_CHANNEL = 'testpilot-telemetry';
```
bwinton: There seems to be a lot of Test-Pilot-specific _and_ SnoozeTabs-specific stuff in here.  How easy would it be to switch this over to use Google Analytics if it ends up graduating into a Shield study?  Would it make sense to split the Test-Pilot-specific stuff out into a separate file that the SnoozeTabs-specific stuff could call, and then switch that file out for a Google-Analytics-specific implementation?

test/.setup.js
```diff
@@ -0,0 +1,24 @@
...
+  if (typeof global[property] === 'undefined') {
+    exposedProperties.push(property);
+    global[property] = document.defaultView[property];
+  }
+});
+
+global.navigator = {
+  userAgent: 'node.js'
+};
+*/
```
bwinton: Do we need this file yet?  Should we not add it until we do?  🙂

src/background.js
lmorchard: Basically, it's dependency injection, so that the tests don't have to rely on a `window` global to pass in a `BroadcastChannel` class because it's not an API based on an importable module

test/.setup.js
lmorchard: We'll need it the first time any React components are tested, and it's handy for setting up globals like `window` when they can't be avoided by doing a dependency injection thing

src/lib/metrics.js
lmorchard: What I've got here now is just basically copying what the Tracking Protection experiment does now. But for Telemetry vs GA (or both), these are all things that will hopefully be addressed with [the pending metrics broker solution from Test Pilot](https://github.com/mozilla/testpilot/issues/1735). Not sure what that will look like yet, though

src/lib/components/MainPanel.js
lmorchard: Yeah, can revise for that. There wasn't a constant yet when I wrote this bit

lmorchard:
Updated to remove `test/.setup.js` and to use the `PICK_TIME` constant in relevant spots.

src/background.js
lmorchard: Not sure - updates weren't listed as one of the measurements in [the plan so far](https://github.com/bwinton/SnoozeTabs/issues/45). Seems like a handy thing to measure, though

docs/metrics.md
```diff
@@ -0,0 +1,113 @@
...
+### Do people use this?
+
+What is the overall engagement of Snooze Tabs? **This is the standard Daily
+Active User (DAU) and Monthly Active User (MAU) analysis.** This captures data
+from the people who have the add-on installed, regardless of whether they are
+actively interacting with it.
+
+### Immediate Questions
+
+TBD
```
lmorchard: Also, need to formulate the questions here. Hoping we have them to be copypasta'd from a product doc somewhere. Or, I can gin them up myself if necessary because I know the best words

package.json
```diff
@@ -49,25 +53,27 @@
     "url": "git+https://github.com/bwinton/SnoozeTabs.git"
   },
   "scripts": {
-    "start": "npm run build && npm-run-all --parallel watch run storybook",
+    "start": "npm run build && npm-run-all --parallel test watch run storybook",
     "start-win": "npm run build && npm run watch",
```
bwinton: Let's add the test to `start-win`, too… (and change it to `npm-run-all --parallel` while we're at it).

src/background.js
bwinton: 👍

src/lib/metrics.js
bwinton: Sounds reasonable.

src/background.js
bwinton: Well, I would say let's do it, then, if we can.

docs/metrics.md
wresuolc: This is from the Product Vision:
> How will we know we are successful?  Key metric here will be retained users. At what rate do experimenters snooze their tabs? Do users snooze tabs with some consistency or does the novelty wear of quickly?
but they could probably use some word tweaking.  /cc @johngruen @mattgrimes @sharonbautista

docs/metrics.md
wresuolc: Seems nice and simple.  r? @6a68 and @sunahsuh

lmorchard:
Rebased against master and added another ping for [when a snoozed tab is updated](https://github.com/bwinton/SnoozeTabs/pull/95/files#diff-683a405c5c005542f9e848b827fe0765R92).

lmorchard:
Considering merging this as-is, since it's big & bit-rotty but does actually work. Pending updates are on changes to semi-unrelated external things that we can roll in as separate issues

bwinton:
Given the rot, I'm 👍 on merging it and filing followup bugs for the other stuff.  (Let me know when the conflict is fixed, and I'll do the merge if you'ld like. 🙂)

docs/metrics.md
johngruen: The basic question we seek to answer is: does snooze tabs provide a useful tool for its users in Test Pilot.

* Do people snooze tabs and at what rate? 
	* x = time, y = number of snoozes (vanity metric)
	* x = time, y = snoozes per user
* Do people close returned tabs without focusing them and at what rate?
	* x= time, y = snoozes focused and snoozes dismissed w/o focus
* Are the measurably different rates of returned tab engagement for snoozes of different lengths.
	* bar graph showing rate at which different snooze length tabs are focused and closed (should show custom times as well (these can be clustered into one group)
* What is the overall distribution of snooze times across all users?
	* Bar graph showing distribution of snooze times
* Do people set custom times, and at what rate, 
	* x = time, y = number of custom snooze events
* Do people re-snooze previously snoozed tabs?
	* x = time, y = snoozes of previously snoozed tabs per user

docs/metrics.md
```diff
@@ -0,0 +1,122 @@
...
+
+### Follow-up Questions
```
wresuolc: we can drop this section

lmorchard:
Rebased, updated `docs/metrics.md`. Think I'll merge this and then file an issue to switch over to [testpilot-metrics](https://github.com/mozilla/testpilot-metrics) ASAP

lmorchard:
Now that this is merged, anyone listening should feel free to submit issues / PRs to further refine `docs/metrics.md`

PR #100, opened 2017-01-17
Run ESLint against all the JS files

pdehaan:
This will also make sure the /stories/* are linted as well as /*.js files.

¯\\\_(ツ)\_/¯

### DIFF
--- a/stories/index.js
+++ b/stories/index.js
@@ -3,8 +3,6 @@ import { storiesOf, action, linkTo } from '@kadira/storybook';
 import { host } from 'storybook-host';
 
 import SnoozePopup from '../src/lib/components/SnoozePopup';
-import MainPanel from '../src/lib/components/MainPanel';
-import ManagePanel from '../src/lib/components/ManagePanel';
 
 // TODO: Get sass working with storybook
 // import '../src/popup/snooze.scss';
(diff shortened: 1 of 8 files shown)


stories/index.js
```diff
@@ -3,8 +3,8 @@ import { storiesOf, action, linkTo } from '@kadira/storybook';
 import { host } from 'storybook-host';
 
 import SnoozePopup from '../src/lib/components/SnoozePopup';
-import MainPanel from '../src/lib/components/MainPanel';
-import ManagePanel from '../src/lib/components/ManagePanel';
+// import MainPanel from '../src/lib/components/MainPanel';
+// import ManagePanel from '../src/lib/components/ManagePanel';
```
bwinton: Why are these commented out (instead of just removed)?

pdehaan: I flipped a coin. I wasn't sure if you were planning on adding more stories or tests for this, and that's why they were imported initially.

But it's easy enough for me to amend the commit and delete the two lines. Your call.

bwinton: I'll merge whatever you have up at the end of this meeting.  Either would be fine.  🙂

pdehaan: Deleted and force pushed. We can always dig through Git history if we need to recreate those two import statements. xoxo

PR #104, opened 2017-01-17
Rename SVGs and optimize all SVGs during build

### LINKED ISSUE #75: Optimize all SVGs and convert filenames to lowercase
Looks like all the SVGs were created in Illustrator 19. We should probably run them all through SVGO or something and optimize them and shave off a few bytes.

Also, I noticed that the filenames are somewhat [consistently inconsistent](https://github.com/bwinton/SnoozeTabs/tree/master/src/icons) where it's either "nightly.svg" or filenames have spaces and mixed case, like "Bell Icon.svg". Probably best to just convert everything to a friendlier bellicon.svg so we don't have any case sensitive, cross platform issues in case we incorrectly tweak some SCSS to "bell icon.svg" or whatever.

pdehaan:
Fixes #75


package.json
```diff
@@ -48,25 +49,25 @@
...
   "scripts": {
-    "start": "npm run build && npm-run-all --parallel watch run storybook",
-    "start-win": "npm run build && npm run watch",
-    "build": "npm run clean && npm-run-all --parallel copy:assets bundle:js bundle:css",
-    "run": "web-ext run -s dist --firefox=firefoxdeveloperedition",
-    "clean": "rm -rf dist && mkdir dist",
-    "watch": "npm-run-all --parallel watch:js watch:css watch:assets",
-    "copy:assets": "shx cp -r src/manifest.json src/icons dist/ && shx mkdir -p dist/popup && shx cp -r src/popup/*.html dist/popup/",
-    "bundle:js": "npm run lint:js && rollup -c --environment entry:background && rollup -c --environment entry:popup/snooze-content",
+    "build": "npm run clean && npm-run-all --parallel copy:assets bundle:*",
```
lmorchard: Alphabetizing the tasks here kind of obscures the PR - not sure what the actual change was for SVG optimization. Also, the order was vaguely intentional as I was trying to keep the more important commands first in the list and roughly sort the sub-commands later in the list.

More a nit than anything, but I'd like to separate things like this into different commits

package.json
```diff
@@ -48,25 +49,25 @@
...
+    "build-storybook": "build-storybook",
     "bundle:css": "npm run lint:sass && shx mkdir -p dist/popup && node-sass src/popup/snooze.scss > dist/popup/snooze.css",
-    "package": "npm run build && web-ext build -s dist -a .",
-    "lint": "npm-run-all --parallel lint:js lint:sass",
+    "bundle:js": "npm run lint:js && rollup -c --environment entry:background && rollup -c --environment entry:popup/snooze-content",
+    "clean": "shx rm -rf dist && mkdir -p dist/popup",
```
bwinton: This needs to call `shx mkdir` instead of `mkdir`, because cygwin's `mkdir` doesn't support the `-p` option.  😭 
(And while we're at it, we don't need the `shx` in front of the `rm -rf`…)

lmorchard: Oh yeah, missed that part. :/ I wonder if CircleCI supports running this stuff under Windows? Would be nice to have a bot do the work for us

package.json
pdehaan: Unsorted for your diff-ing pleasure.

package.json
pdehaan: Oh Cygwin... `shx` seems tricky, but probably a better solution than having to resort to `rimraf` and a handful of other libs for cross platform greatness.

PR #111, opened 2017-01-19
fix: Disable "Snooze" button when selected time passes

### LINKED ISSUE #73: Unable to snooze a tab to a date/time in the past
Super duper edge case, but oddly SnoozeTabs doesn't try and stop me...

### Steps to reproduce:
1. Go to a page which we want to hypothetically snooze. For example, https://www.mozilla.org/en-US/firefox/new/
1. Open the Browser Console and make sure any previous logs/errors are cleared (for easier debugging).
1. Snooze the tab but pick a specific Date/Time. For example, when selecting a Date, hit "Now" and it should select the current date/time, and disable the "Snooze" button.
1. Click the "Select Time" and hit "Now" button a few more times, and the time picker should scroll to the exact current date/time again and still not let you select the current time.
1. Select a second value a few seconds in the future so the "Snooze" button becomes enabled.
1. Wait a few more seconds so the selected date/time is now _just_ technically in the past (but the "Snooze" button is still enabled).
1. Hit the enabled "Snooze" button.

### Actual results
Nothing happens, and nothing seems to appear in my Developer Tools or Browser Console informing me there was an error.
In order to _actually_ snooze the active tab, I need to quickly select a date a few seconds in the future and then click the Snooze button again while that date is still in the future.

Not sure if we should be aggressively preventing users from snoozing tabs in the past, or if we just shrug it off and then display the snooze tab reminder immediately.

Pretty edge case.

lmorchard:
Fixes #73.

### DIFF
--- a/src/lib/components/DatePickerPanel.js
+++ b/src/lib/components/DatePickerPanel.js
@@ -6,13 +6,26 @@ import classnames from 'classnames';
 import Calendar from 'rc-calendar';
 import TimePickerPanel from 'rc-time-picker/lib/Panel';
 
+// Arbitrary 0.5s interval for live validation of time selection
+const VALIDATION_INTERVAL = 500;
+
 export default class DatePickerPanel extends React.Component {
   constructor(props) {
     super(props);
+    this.validationTimer = null;
     this.state = {
       currentValue: props.defaultValue,
       confirmDisabled: false
     };
+  }
+
+  componentDidMount() {
+    this.validationTimer = setInterval(this.validateOnTimeChange.bind(this),
+                                       VALIDATION_INTERVAL);
+  }
+
+  componentWillUnmount() {
+    clearInterval(this.validationTimer);
   }
 
   componentWillReceiveProps(nextProps) {
@@ -100,6 +113,14 @@ export default class DatePickerPanel extends React.Component {
       currentValue: value,
       confirmDisabled: (value.valueOf() <= Date.now())
     });
+  }
+
+  validateOnTimeChange() {
+    const { currentValue, confirmDisabled } = this.state;
+    const newConfirmDisabled = (currentValue.valueOf() <= Date.now());
+    if (confirmDisabled !== newConfirmDisabled) {
+      this.setState({ confirmDisabled: newConfirmDisabled });
+    }
   }
 
   handleConfirm() {


src/lib/components/DatePickerPanel.js
```diff
@@ -6,15 +6,28 @@ import classnames from 'classnames';
 import Calendar from 'rc-calendar';
 import TimePickerPanel from 'rc-time-picker/lib/Panel';
 
+// Arbitrary 0.5s interval for live validation of time selection
+const VALIDATION_INTERVAL = 500;
```
bwinton: Given that we don't show the seconds anymore, what do you think about changing this to, like, 10s, or 30s, or even 60s?  🙂

src/lib/components/DatePickerPanel.js
```diff
@@ -102,6 +115,14 @@ export default class DatePickerPanel extends React.Component {
     });
   }
 
+  validateOnTimeChange() {
+    const { currentValue, confirmDisabled } = this.state;
+    const newConfirmDisabled = (currentValue.valueOf() <= Date.now());
+    if (confirmDisabled !== newConfirmDisabled) {
+      this.setState({ confirmDisabled: newConfirmDisabled });
```
bwinton: I _thought_ `setState` would take care of figuring out whether they were different or not.  Do we need to double-check that ourselves, too?  Is it much faster if we just don't call `setState`?

src/lib/components/DatePickerPanel.js
```diff
@@ -102,6 +115,14 @@ export default class DatePickerPanel extends React.Component {
     });
   }
 
+  validateOnTimeChange() {
+    const { currentValue, confirmDisabled } = this.state;
+    const newConfirmDisabled = (currentValue.valueOf() <= Date.now());
```
bwinton: Maybe we should check that the current value is less than five seconds from now, to give ourselves a bit of a buffer?

src/lib/components/DatePickerPanel.js
lmorchard: No, I think React is lazy here - if you call `setState` an update gets queued. That state update object could have some deep structured data, so I think they leave it up to the component to decide.

PR #117, opened 2017-01-23
fix: Sync snoozed tabs to a bookmark folder.

### LINKED ISSUE #42: Save snoozed tabs somewhere on uninstall
We've been talking about skipping full sync to bookmarks with snoozed tabs. But, we should still have a bookmarks folder to which snoozed tabs can be exported on extension uninstall and imported on install. Maybe also as a manual action?

bwinton:
We can't get an event on uninstall, so let's just always keep them around.
They _should_ be in the same order as in the Manage panel.

Fixes #42.

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -76,9 +76,7 @@ function init() {
       updated[item[0]] = item[1];
     });
     log('setting next open tabs to now', updated);
-    return browser.storage.local.set(updated).then(() => {
-      updateWakeAlarm();
-    });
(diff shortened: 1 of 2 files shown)


pdehaan:
Working for me...

firefox_developer_edition_start_page_and_mozilla_tippy-top-sites
**Figure 1:** testing w/ today.com, tomorrow.com, thisweekend.com, and nextweek.com. Correctly sorted in the Bookmarks folder by their expiration rather than their name.

firefox_developer_edition_start_page_and_library

library

PR #120, opened 2017-01-24
fix: Rework all styling to better match UX spec

### LINKED ISSUE #68: Calendar overlay should not take up full-width of the panel
Currently the calendar panel overlay looks something like this:

screen shot 2017-01-06 at 3 12 11 pm

This panel should not take up the full width of the Snooze Tabs panel:

image

This is to match how overlay panels work elsewhere in Firefox. It provides some context for the user on how they got there, and where they can go back to. Basically the background should grey out slightly as the panel slides out. See History or Synced Tabs when the icon is in the hamburger menu.

### LINKED ISSUE #101: Style calendar picker
Ref: #11

### LINKED ISSUE #107: VIsual tweaks to panel
Just a few visual tweaks to better match the current styling to the spec.

screen shot 2017-01-18 at 4 16 20 pm

### LINKED ISSUE #113: [Enhancement] Hover state is missing from the "Manage Snoozed Tabs" and "Back" buttons
[Affected versions]:
- Nightly 53.0a1
- Snooze Tabs 1.0 (built on 1/20/2017)

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- Have a Firefox profile with the latest Snooze Tabs add-on version (1.0-dev, built on 1/20/2017) loaded as a temporary add-on.

[Steps to reproduce]:
1. Open the browser with the profile from prerequisite and navigate to any website.
2. Click the "Snooze Tabs" icon from the main browser toolbar.
3. Hover the "Manage Snoozed Tabs" button.
4. Click on the "Manage Snoozed Tabs" button.
5. Hover the "Back" button.

[Expected result]:
- When hovering the "Manage Snoozed Tabs" or the "Back" button, it gains a darker overlay.

[Actual result]:
- When hovering the "Manage Snoozed Tabs" or the "Back" button, it remains unchanged.

[Notes]:
- The same behavior occurs for "Pick a date " section bottom part buttons.
- Attached a screen recording of the issue:
snoozetabs_hoverissue

### LINKED ISSUE #118: Need prettier blank state for manage tabs
```sh
$ git rev-parse --short HEAD # dc9b49d
```

### Steps to reproduce:
```sh
$ nvm use 6
$ npm install
$ npm start
```

### Actual results:
firefox_developer_edition_and_new_issue_ _bwinton_snoozetabs

### Expected results:
My browser is launched with a "New Tab" experience.
Clicking the "SnoozeTabs" button brings up an empty **Manage Snoozed Tabs** panel, since this is first run and the first time I've interacted with the button.

lmorchard:
* Styling tweaks across all panels to better match UX spec
* Calendar styling reworked, "now" link removed
* Switch to inline time picker instead of separate view
* Time display formatting tweaks
* Sub-panels no longer full-width and dimming effect added underneath
* Add Fira Sans web font
* Additional storybook story to reflect non-snoozable tab version of
  management panel

Fixes #68.
Fixes #101.
Fixes #107.
Fixes #113.
Fixes #118.


lmorchard:
Probably some more tweaking to do from here, but figured this is a big enough chunk of things to start getting some looks at it

lmorchard:
Also will roll these colors in: https://github.com/bwinton/SnoozeTabs/issues/113#issuecomment-274894812

lmorchard:
Okay, should include rollover & active colors for footer buttons now

src/popup/snooze.scss
```diff
@@ -1,34 +1,80 @@
...
+  // margin: 0 0 50px 0;
+  margin: 0;
   -moz-user-select: none;
   user-select: none;
 }
 
+.panel-wrapper {
+  width: 100%;
+  height: 100%;
+  background-color: #333333;
```
bwinton: Maybe we should put the `background-color` on the `body`, so that the arrow will pick it up…

lmorchard: Well, the problem is that the arrow is then always a dark color. This bg color is only there to show through when turning down the opacity for a background panel

lmorchard:
Worth noting that these styles are now pretty horrible when the icon is moved into the hamburger menu and comes up extra narrow. 

Does it make sense to merge this chunk and then circle back for more tweaks? Or go back to the drawing board on this PR?

bwinton:
I think we could massage this PR into a reasonable state for the menu panel case.  We also need to see what Sevaan asks for…

src/popup/snooze.scss
bwinton: Hmm.  On Windows, the arrow is already a dark colour…

lmorchard:
Is there a way to detect the width of the panel? Wondering if I can look at the body width or something to slap a className on and pivot styles based on that

bwinton:
Media query maybe?
```css
@media (max-width: 200px) {
  .panel-wrapper {
    display: none;
  }
}
```
😉

pdehaan:
When managing snoozes, I noticed that the horizontal separator lines don't always extend 100% width.

Plus, the little doorhanger arrow looks black instead of white.

menubar_and_nightly_start_page_and__enhancement__hover_state_is_missing_from_the__manage_snoozed_tabs__and__back__buttons_ _issue__113_ _bwinton_snoozetabs

lmorchard:
> When managing snoozes, I noticed that the horizontal separator lines don't always extend 100% width.

They're not supposed to per the UX spec in #107

lmorchard:
> Media query maybe?

Oof, I always forget about those. Also seems like they don't play well with the storybook thing because the media query isn't triggered by the component inclusion by the page at large. That makes tweaking styles extra fun. Ugh. Will keep poking at it a bit though

lmorchard:
Okay, I think I have [a React hack](https://github.com/lmorchard/SnoozeTabs/commit/c4e46e0d6631a1ad83c1be4b7ae6d5ef7a268ca3) that works both in the storybook and in the webextension that I'll try running with. 

Media queries work in the extension but not the storybook. But, the storybook is the easiest place to iterate on styles because I can't seem to keep the Browser Toolbox inspector working for more than a CSS tweak or two on the real extension.

lmorchard:
Also likely in need of tweaks, but I went through and just kind of fudged font sizes and paddings to make things not as bad under the hamburger menu.

bwinton:
Something went kinda wrong with the new push.  I now don't see the popup at all on Developer Edition…

lmorchard:
Crud, now it works in the hamburger menu but not in the popup. Still wrestling with it & webext weirdnesses

lmorchard:
Okay... I think *this* should work better. Figured out that while content determines the size of the popup from the toolbar, the width is fixed for the menu panel. So, I added a resize event listener on the window that can detect that from `document.body` and toggle a state prop in the SnoozePanel component to toggle styles. Works in the Storybook as a bonus!

src/lib/components/MainPanel.js
```diff
@@ -45,7 +45,7 @@ export default class MainPanel extends React.Component {
       <li className="option" key={item.id} id={item.id} onClick={ ev => this.handleOptionClick(ev, item) }>
         <img src={ `../icons/${item.icon || 'nightly.svg'}` } className="icon" />
         <div className="title">{item.title || '&nbsp;'}</div>
-        <div className="date">{date}</div>
+        <div className="date">&nbsp;{date}</div>
```
bwinton: Why the `&nbsp;` here?

src/popup/snooze-content.js
```diff
@@ -10,11 +10,15 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import SnoozePopup from '../lib/components/SnoozePopup';
 
+// HACK: Arbitrary value found by measuring menu panel popup width
```
bwinton: Hah!  😁

src/popup/snooze.scss
```diff
@@ -254,53 +357,127 @@ li.option {
   color: #6E6E6E;
 }
 
-.option > .date {
-  flex: 0 0 65px;
+.rc-calendar {
+  font-family: 'Fira Sans';
+  font-size: 14px;
+  letter-spacing: -.69px;
```
bwinton: We seem to use these three lines a few times.  Should we pull them out into a mixin?  (Maybe in a future patch…)

src/popup/snooze.scss
```diff
@@ -1,34 +1,82 @@
 @import 'node_modules/rc-calendar/assets/index';
 @import 'node_modules/rc-time-picker/assets/index';
 
-$panel-width: 250px;
+$panel-width: 320px;
+$panel-height: 474px;
+$panel-shadow-width: 13px;
+
+$narrow-panel-width: 230px;
+$narrow-panel-height: 550px;
```
bwinton: I think this depends on how much stuff you have in your menu panel…  (Try adding a ton of buttons to it, and see what happens.)

src/lib/components/MainPanel.js
lmorchard: Otherwise, the chevron for "Pick a date/time" doesn't seem to show up

src/popup/snooze.scss
lmorchard: Hmm, maybe this should just be 100%?

sevaan:
The Pick a Date/Time row has rounded corners at the bottom which should be removed.

screen shot 2017-02-01 at 11 39 22 am

sevaan:
If you are on a page you cannot snooze (e.g. about:newtab), when you click on the panel the Manage Snoozed Tabs panel should open. Currently there is a big empty blank space at the bottom of the panel. Ideally the panel height should collapse to show the relevant tabs if there are any. If this isn't possible, the dark space should be white. If there are no tabs to manage, show the empty Manage Tabs state (Issue #118)

screen shot 2017-02-01 at 11 41 42 am

Also, when clicking on a tab, it opens in a new active tab, but the panel remains open. The panel should close if the user has selected a tab to open.

sevaan:
If I repeatedly open the Snooze Tabs panel, the width of the panel is sometimes correct and sometimes skinny.

panel width

sevaan:
If I repeatedly open the Snooze Tabs panel, the width of the panel is sometimes correct and sometimes skinny.

panel width

src/lib/components/MainPanel.js
bwinton: Okay, cool.

src/popup/snooze.scss
bwinton: Yeah, that seems reasonable.  Give it a try, lemme know.  :smile:

lmorchard:
Another try... I think I got most of the issues addressed. The case where it's in the menu panel is still a bit flakey, and I can't quite get the dimming right in that case when a subpanel slides over. Kind of want to get this merged and keep refining though. There's a lot in this PR now, covering 5 issues.

lmorchard:
Hmm, actually, this is starting to look like it blocks other things while it sits in the air. So I think I'll just merge it and we can fix going forward

PR #125, opened 2017-01-26
add guid to schema

wresuolc:


lmorchard:
What is this guid? Where's it come from?

wresuolc:
comes out of telemetry automatically.  it's already being sent, this is just recording it along with our other data so we can answer funnel questions

lmorchard:
Cool beans, merging!

PR #128, opened 2017-01-31
fix: Add a confirmation when snoozing a tab.

### LINKED ISSUE #60: Add tab-close confirmation modal
After a user clicks a snooze option, we want to inform them that the tab is going to be closed, as well a provide the option to not see the confirmation modal in the future.

UX: https://mozilla.invisionapp.com/share/F29T5JA8S#/211984424_Today_Confirmation

Today: "This tab will close and be reopened today at _time_"
Tomorrow: "This tab will close and be reopened tomorrow at _time_"
This Weekend: "This tab will close and be reopened on _date_ at _time_"
Next Week: "This tab will close and be reopened on _date_ at _time_"
Next Month: "This tab will close and be reopened on _date_ at _time_"
Custom Date/Time: "This tab will close and be reopened on _date_ at _time_"

bwinton:
So much stuff.  I really should have done separate commits.  😞

* Add a confirm-bar page-script.
* Load the close button svg for the confirm-bar page-script.
* Add a toggle to never show the confirm-bar again,
  * and save it in local storage,
  * and filter it out of the list of entries.
* Wait for a confirmation before snoozing the tab.
* Close the tab from the background script, instead of from the popup.

We would like the time text to be a way to let the user edit the time, but that seemed too hard for this PR, so I'm going to try and leave it for a followup.  (Also I'm going to have to fix this up quite a bit once your giant styling PR lands, Les. 🙂)

Fixes #60.


src/background.js
```diff
@@ -106,6 +139,9 @@ const messageOps = {
     Metrics.updateSnoozedTab(message);
     return messageOps.cancel(message.old).then(() => messageOps.schedule(message.updated));
   },
+  setconfirm: message => {
+    browser.storage.local.set({dontShow:message.dontShow})
```
pdehaan:
```sh
/home/ubuntu/SnoozeTabs/src/background.js
  143:59  error  Missing semicolon  semi

✖ 1 problem (1 error, 0 warnings)
```

bwinton: Yeah, I couldn't figure out why that wasn't fixed on my following two pushes.  Turns out you need to `git add` before your `git commit`.  And on that note, I think I should stop writing code for the day. 😉

package.json
```diff
@@ -60,7 +60,7 @@
     "clean": "rm -rf dist && shx mkdir -p dist/popup",
     "watch": "npm-run-all --parallel watch:*",
     "copy:assets": "shx cp -r src/manifest.json src/icons dist/ && svgo src/icons -o dist/icons && shx cp -r src/popup/*.html dist/popup/",
-    "bundle:js": "npm run lint:js && rollup -c --environment entry:background && rollup -c --environment entry:popup/snooze-content",
+    "bundle:js": "npm run lint:js && rollup -c --environment entry:background && rollup -c --environment entry:popup/snooze-content && rollup -c --environment entry:lib/confirm-bar",
```
pdehaan: I've never used rollup before, but not sure if we could use something like [**rollup-plugin-multi-entry**](https://github.com/rollup/rollup-plugin-multi-entry) to specify an array of entry points instead of calling the full `rollup -c --environment entry:*` command 3+ times.

src/lib/confirm-bar.js
```diff
@@ -0,0 +1,126 @@
...
+        flex: 1;
+      }
+      #${closeId} {
+        cursor: pointer;
+        height: 10px;
+        margin-right: 16px;
+        width: 10px;
+      }
+    </style>
+    <img src="${iconData}">
```
pdehaan: Not sure if these `` tags need `alt=""` tags for accessibility.

pdehaan:
Need to do some more testing on dates and how I got into this state, but looks like we have an edge somewhere...

cnn_-_breaking_news__latest_news_and_videos_and__1___opengifs___freenode_and_snoozetabs_rollup_config_js_at_master_ _bwinton_snoozetabs

pdehaan:
Minor edge case where I have only one tab open...

### Steps to reproduce:
1. Go to http://www.cbs.com (no other tabs open).
1. Snooze tab until "Later today".
1. Click on Snooze tabs toolbar icon and notice that nothing was snoozed.

### Actual results:
Because I snoozed my only open tab, a new tab was opened [and focused] in my browser, the tab with the confirmation panel got pushed into the background and was never fully snoozed. Could potentially cause some people mistakenly assuming their tab was snoozed and lost.

---

nightly_start_page

sevaan:
Can we wrap the text down, and push the height of the bar down to accommodate it for when the browser width is small?

screen shot 2017-02-01 at 12 36 40 pm

sevaan:
Let's add hover/press colours to the buttons:

**Blue Button**
Hover: #0670cc
Press: #005bab

**White Button**
Hover: #ebebeb
Press: #d4d4d4

src/background.js
```diff
@@ -66,11 +67,22 @@ function init() {
...
+      return response.arrayBuffer();
+    }).then(function(response) {
+      closeData = 'data:image/svg+xml;base64,' + btoa(String.fromCharCode(...new Uint8Array(response)));
+    }).catch(reason => {
+      log('init get closeData rejected', reason);
     });
   }
 
   browser.storage.local.get().then(items => {
+    delete items.dontShow;
```
lmorchard: Hmm, this is the sort of thing I'm getting snagged on for wrapping up the metrics PR. Need to generate & persist a per-client UUID, and didn't see how to do that without reworking the whole storage scheme to move the timers into their own property rather than using the "root" of storage.

Maybe I can just kind of generalize this "delete" thing and have bits that look for alarms filter out known non-alarm keys? (e.g. dontShow, metricsUUID, whatever else might come up)

lmorchard:
So I tore the bandaid off on #120 - and yeah I broke this PR :( But, playing with the branch locally, things look pretty good. I'll probably want to generalize the whole "things other than alarms in storage" thing over in my metrics branch after this, though

lmorchard:
The conflicts don't look *too* bad, but I'm a little leery to use the "Resolve conflicts" button here

package.json
bwinton: We can't, because we want multiple outputs.  :(

src/lib/confirm-bar.js
bwinton: Fixed.

bwinton:
Okay, I _think_ I got all the comments, but a re-review would be nice to make sure.  🙂

sevaan:
It's looking good.

Let's change it from "Don't Show this again" to be a checked box with the string "Ask me every time".

sevaan:
When snoozing a custom time, say "12:53pm" the confirmation notification bar shows "12pm". We should be displaying the exact time the user has picked.

The preset options (e.g. Later Today, Tomorrow, etc.) show rounded times...so if it is 12:10pm, it would say it it coming back at 3pm (+3 hours and rounded). The tab should actually come back at that rounded time, not at 3:10pm.

PR #136, opened 2017-02-02
feat: Switch to new testpilot-metrics broker library for metrics pings

lmorchard:


lmorchard:
This collides a little with #128, at least insofar as the `dontShow` storage property goes. Trying to pull some of that into its own module, though, as I add another storage property for `metricsUUID`

test/lib/storage-test.js
```diff
@@ -0,0 +1,73 @@
+/* global describe, beforeEach, it */
```
pdehaan: Instead of adding `/* global ... */` in all the ./test/*.js files, we may just want to create a ./test/.eslintrc.json file with the following sauce:

```json
{
  "env": {
    "mocha": true
  },
  "rules": {
    "promise/avoid-new": "off",
    "promise/no-callback-in-promise": "off"
  }
}
```

Plus, then we can remove all the rogue `// eslint-disable-line ...` comments below.

src/lib/storage.js
```diff
@@ -0,0 +1,32 @@
+/* global browser */
```
pdehaan: nit: You shouldn't need `/* global browser */` here since we define the `browser: true` env in the root .eslintrc.json file.

package.json
```diff
@@ -52,14 +53,18 @@
...
   "scripts": {
     "start": "npm run build && npm-run-all --parallel test watch run storybook",
     "start-win": "npm run build && npm-run-all --parallel test watch",
-    "build": "npm run clean && npm-run-all --parallel copy:assets bundle:*",
+    "build": "npm run clean && npm-run-all --parallel copy:* bundle:*",
     "run": "web-ext run -s dist --firefox=nightly",
     "clean": "rm -rf dist && shx mkdir -p dist/popup",
     "watch": "npm-run-all --parallel watch:*",
     "copy:assets": "shx cp -r src/manifest.json src/icons dist/ && svgo src/icons -o dist/icons && shx cp -r src/popup/*.html src/popup/FiraSans-Regular.* dist/popup/",
+    "copy:js": "shx cp node_modules/testpilot-metrics/testpilot-metrics.js dist/",
```
lmorchard: Not super happy with copying this over as a script, rather than importing it as a module. But, testpilot-metrics does some conditional `require()` stuff to work for WebExtensions, SDK add-ons, and Test Pilot itself - and rollup doesn't like that.

lmorchard:
Rebased against master, made some tweaks per @pdehaan comments

src/lib/metrics.js
```diff
@@ -1,32 +1,51 @@
+/* global Metrics */
```
lmorchard: This `Metrics` comes from testpilot-metrics, included as a script alongside `background.js` in the `manifest.json`

lmorchard:
Oh yeah, and though we don't have Telemetry stuff hooked up for these pings yet, the GA_TRACKING_ID leads to an actual Google Analytics instance I set up on my personal account (at least for now) that I can invite interested parties to check out

lmorchard:
It's a little hacky, but rebased against master and resolved some conflicts with the `dontShow` storage property. Didn't completely move all storage stuff into that module though

lmorchard:
I might have gone a little overboard with `src/lib/storage.js`, but it should be easier to tweak how we do storage in the future if/when we want to add more properties or change the storage layout

package.json
pdehaan: Ref: https://github.com/mozilla/testpilot-metrics/issues/16 &mdash; this _may_ break in the future if **testpilot-metrics** splits up into multiple files.

lmorchard: Not too worried about that right now, since that will likely be a major version upgrade if/when that happens.

src/lib/storage.js
```diff
@@ -0,0 +1,65 @@
...
+export const KEY_DONT_SHOW = 'dontShow';
+
+export const KNOWN_PROPERTIES = [
+  KEY_DONT_SHOW,
+  KEY_METRICS_UUID
+];
+
+export function getAlarmsAndProperties() {
+  return browser.storage.local.get().then(raw => Object.keys(raw)
+    .reduce((obj, key) => {
```
bwinton: Ewwww.  (I'm not a huge fan of `reduce`.  But it works, so let's go with it. 😉)

lmorchard:
Think I'll merge this after CircleCI wraps up, should unblock us for security review

Issue #142, opened 2017-02-03
Also, when button is moved in the Firefox menu, the gray margin does not cover the entire panel.


lmorchard:
The "shadow" of the management panel over the main snooze panel ends before it reaches the bottom of the panel.

<img width="297" alt="screenshot 2017-02-03 13 37 20" src="https://cloud.githubusercontent.com/assets/21687/22603565/f21599c6-ea15-11e6-93a2-4b09a0cdbf1d.png">

Having trouble getting the CSS to cover the full height of the panel in the menu case.

PR #143, opened 2017-02-03
fix: More narrow panel style tweaks for Windows

### LINKED ISSUE #137: Items become unreadable or even unreachable because of doorhanger resize, when "SnoozeTabs" button is moved in the Firefox menu
[Notes]:
This issue causes the next problems
- The Snooze options are not entirely visible.
- The items from the "Managed Snoozed Tabs" panel can't be removed.
- The calendar for "Pick a Date/Time" snoozing option is not entirely visible, making the Saturday days impossible to be selected.

[Affected versions]:
- Nightly 54.0a1
- Snooze Tabs 1.0.9 (built on 02/02/2017)

[Affected Platforms]:
- All Windows

[Prerequisites]:
- Have a Firefox profile with the latest Snooze Tabs add-on version (1.0.9 -dev, built on 02/02/2017) loaded as a temporary add-on.

[Steps to reproduce]:
1. Open the browser with the profile from prerequisite and navigate to any website.
2. Click the "SnoozeTabs" button and observe the doorhanger size and items.
3. Right-click on the Snooze Tabs icon from the toolbar and select the "Move to Menu" option from the doorhanger.
4. Click the Firefox menu from the main toolbar.
5. Click on the "SnoozedTabs" icon and observe the doorhanger size and items.

[Expected result]:
- The "SnoozedTabs" doorhanger has the same width as when displayed in toolbar and all items are visible and reachable.

[Actual result]:
- The "SnoozedTabs" doorhanger has a reduced width making some of the items unreadable or even unreachable.

[Additional Notes]:
- The issue isn't reproducible on Mac OS.
- Attached a screen recording of the issue:
snoozetabs_panel_isssue

lmorchard:
Fixes #137.

### DIFF
--- a/src/popup/snooze-content.js
+++ b/src/popup/snooze-content.js
@@ -11,8 +11,10 @@ import ReactDOM from 'react-dom';
 import { getAlarmsAndProperties } from '../lib/storage';
 import SnoozePopup from '../lib/components/SnoozePopup';
 
-// HACK: Arbitrary value found by measuring menu panel popup width
-const NARROW_PANEL_MIN_WIDTH = 225;
+// HACK: Arbitrary breakpoint for styles below which to use "narrow" variant
+// The panel width is specified in Firefox in em units, so it can vary between
+// platforms. OS X is around 224px, Windows is around 248px.
+const NARROW_PANEL_MIN_WIDTH = 275;
 
 const DEBUG = (process.env.NODE_ENV === 'development');
 
--- a/src/popup/snooze.scss
+++ b/src/popup/snooze.scss
@@ -5,10 +5,6 @@ $panel-width: 320px;
 $panel-height: 474px;
 $panel-shadow-width: 13px;
 
-$narrow-panel-width: 225px;
-$narrow-panel-height: 100%;
-$narrow-panel-shadow-width: 7px;
-
 @font-face {
   font-family: 'Fira Sans';
   src: url('FiraSans-Regular.woff') format('woff');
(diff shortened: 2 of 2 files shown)


lmorchard:
Turns out we can't rely on exact pixel widths across platforms for the fixed menu panel size, because that size is [fixed in `em` units](http://searchfox.org/mozilla-central/source/browser/themes/shared/customizableui/panelUI.inc.css#7) and varies across platforms in terms of pixel width.

src/popup/snooze-content.js
```diff
@@ -10,8 +10,10 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import SnoozePopup from '../lib/components/SnoozePopup';
 
-// HACK: Arbitrary value found by measuring menu panel popup width
-const NARROW_PANEL_MIN_WIDTH = 225;
+// HACK: Arbitrary breakpoint for styles below which to use "narrow" variant
+// The panel width is specified in Firefox in em units, so it can vary between
+// platforms. OS X is around 224px, Windows is around 248px.
```
bwinton: Ooh, I wonder what it is on Windows HiDPI mode?  (I'm currently running at 225%, because I'm old…  😉)
I wonder if we should check `Window.devicePixelRatio` too?

bwinton:
(@sevaan: Pssst!  This is waiting for your UI-Review… 😉)

sevaan:
Looking good! Few things:

slice
slice 2
screen shot 2017-02-07 at 12 44 43 pm

lmorchard:
FWIW, there's an issue open for the opacity layer thing (#142). Can we split the rest of these out into separate issues? I don't think most of them are new with / caused by this PR. Also, is this on Mac or Windows?

src/popup/snooze-content.js
lmorchard: Hmm, I only have a 1080p monitor on my Windows machine. Is this breaking for you? Can you see what the console messages say? There should be "resize" messages with the detected pixel width of the panel

lmorchard: FWIW, I switched my machine to 175% - which is as high as it will go. The panel width gets reported as 249px and styles look okay to me

ckprice:
Per checkin, let's break up Sevaan's comment into separate issues and leave out of this PR.

lmorchard:
rebased against master, may merge and see where we can go from here with further PRs

src/popup/snooze-content.js
bwinton: Windows got `resize 248 true true` at 350%, and `resize 176 true true` at 250% (recommended), for what that's worth.

PR #150, opened 2017-02-07
chore: bump version to 1.0.11

lmorchard:
Forgot to do this before production push, so this is kind of a retro-port of the version bump I did there.


lmorchard:
Hmm, looks like the production push didn't quite work. Bumping version again. If that works, I'll self-merge this.

PR #152, opened 2017-02-07
feat: Allow users to edit default snooze times

### LINKED ISSUE #69: Allow users to edit default snooze times
Our user research has shown that users want to customize some of the default times provided in the Snooze Tabs panel list.

For example, currently when a user opens the panel they can snooze a tab until tomorrow morning at 9:am. Some users want to bring it back tomorrow morning, but maybe at 10am instead. Right now they would have to go through the Custom Date/Time option which can be tedious.

Instead, if the user clicks on the row, snooze as usual. If, however, they hover over the date in the row, change the hover colour over the date to #009CFC and make it a hyperlink. If they click the link, slide over the panel with the right date selected and the 10 box of the time selector already selected, so they can quickly edit the time and move on.

screen shot 2017-01-06 at 3 16 48 pm

lmorchard:
- Adds `defaultDateChoice` state to MainPanel

- Style & event handler to turn option dates into links

- Link click pre-fills the datetime picker with the clicked value

Fixes #69.

### DIFF
--- a/src/lib/components/MainPanel.js
+++ b/src/lib/components/MainPanel.js
@@ -11,14 +11,14 @@ export default class MainPanel extends React.Component {
     super(props);
     this.state = {
       datepickerActive: false,
-      dateChoice: null
+      dateChoice: null,
+      defaultDateChoice: props.moment()
     };
   }
 
   render() {
     const { id, active, moment } = this.props;
-    const { datepickerActive } = this.state;
-
+    const { datepickerActive, defaultDateChoice } = this.state;
 
     return (
       <div>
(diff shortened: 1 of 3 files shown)


src/lib/components/MainPanel.js
```diff
@@ -11,13 +11,14 @@ export default class MainPanel extends React.Component {
     super(props);
     this.state = {
       datepickerActive: false,
-      dateChoice: null
+      dateChoice: null,
+      defaultDateChoice: moment()
```
bwinton: We'll need to change this to `props.moment()`, as I suspect you know.
(And there are a couple of other `moment`s down below that will need changing, too.)

lmorchard:
Rebased against master, resolved conflicts.

PR #162, opened 2017-02-09
refactor: Build changes for NODE_ENV={development,production}

### LINKED ISSUE #155: We shouldn't show the "Real Soon Now" item in Production builds.
Ya know, now that we have actual production builds and all…

lmorchard:
- Only include "Real Soon Now" snooze time when `NODE_ENV===development`

- Use real `NODE_ENV` in `rollup.config.js` setup of `replace()` plugin,
  with a default of `development` when not set

- Move to inline conditionals using `process.env.NODE_ENV` because
  rollup will omit code blocks from bundles where that resolves to false

- Build both development & production artifacts in CircleCI

- README updates

Fixes #155.

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -11,13 +11,11 @@ import { NEXT_OPEN, PICK_TIME, times, timeForId } from './lib/times';
 import Metrics from './lib/metrics';
 import { getAlarms, saveAlarms, removeAlarms,
          getMetricsUUID, getDontShow, setDontShow } from './lib/storage';
+import { makeLogger } from './lib/utils';
 
-const DEBUG = (process.env.NODE_ENV === 'development');
+const log = makeLogger('BE');
+
 const WAKE_ALARM_NAME = 'snooze-wake-alarm';
-
-function log(...args) {
-  if (DEBUG) { console.log('SnoozeTabs (BE):', ...args); }  // eslint-disable-line no-console
-}
 
 let iconData;
 let closeData;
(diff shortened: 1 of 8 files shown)


src/background.js
```diff
@@ -12,11 +12,12 @@ import Metrics from './lib/metrics';
...
          getMetricsUUID, getDontShow, setDontShow } from './lib/storage';
 
-const DEBUG = (process.env.NODE_ENV === 'development');
 const WAKE_ALARM_NAME = 'snooze-wake-alarm';
 
 function log(...args) {
-  if (DEBUG) { console.log('SnoozeTabs (BE):', ...args); }  // eslint-disable-line no-console
+  if (process.env.NODE_ENV === 'development') {
+    console.log('SnoozeTabs (BE):', ...args);  // eslint-disable-line no-console
+  }
```
bwinton: So, it's a small function, but we seem to have it in three different places.  What do you think about throwing it in a util file?

lmorchard:
Rebased against master and moved the copypasta'd log function into a utils module. Will merge if/when tests pass

src/lib/utils.js
```diff
@@ -0,0 +1,9 @@
+export function makeLogger(prefix) {
+  if (process.env.NODE_ENV === 'development') {
+    return (...args) => {
+      console.log(`SnoozeTabs (${prefix})`, ...args);  // eslint-disable-line no-console
```
pdehaan: Should this have a `:` after the `(${prefix})` bit? (I think it did previously)

lmorchard: It did, but doesn't really matter... just arbitrary logging chatter, not parsed by anything except eyeballs

PR #168, opened 2017-02-13
feat: Add l10n support!

bwinton:
I strongly suspect there's something “not right” in the way I've chosen to expose some of the strings, so we shouldn't merge this until we hear from our friendly neighbourhood localizer, but if you wanted to review the code, and see how that looks, I'd love to fix whatever I could while we wait.  🙂

The `pontoon-to-webext.js` file was largely copied from [`pontoon-to-json`](https://github.com/mozilla/pontoon-to-json), but modified to do what we needed instead.

~~I also need to think a little about how we're going to localize the dates with `moment.js`, but hopefully that won't conflict much with what's already here.~~ Done!

Oh, and if you can figure out a better way to avoid the test failures in 73484f3, I'd love to not use the hack from 5886cb7!

I think that's all…  🙂


locales/en-US/snoozetabs.properties
```diff
@@ -0,0 +1,45 @@
+extDesc=An add-on to let you snooze your tabs for a while.
```
flodolo: I think the concept of 'snooze' is going to be hard to grasp for some localizers. Can we add a really short comment explaining what the add-on does? That will help them find a good translation.

To give you an idea, there's no perfect translation of 'snooze' in Italian, it's commonly translated as 'pospone'.

locales/en-US/snoozetabs.properties
```diff
@@ -0,0 +1,45 @@
...
+popupTitle=Snooze tab until…
+contextMenuTitle=Snooze tab until…
+bookmarkFolderTitle=Snoozed Tabs
+notificationTitle=Tab woke up…
+
+confirmIconAltText=tab icon
+# LOCALIZATION NOTE: placeholder is the time the tab will snooze until, as per https://github.com/bwinton/SnoozeTabs/issues/60
```
flodolo: Actually reference $TIME$ instead of placeholder in the comment

locales/en-US/snoozetabs.properties
```diff
@@ -0,0 +1,45 @@
...
+confirmTimeTitle=This tab will snooze until $TIME$.
```
flodolo: Per IRC discussion: if we could use a format localizers are used to, like `{time}`, it would probably reduce the risk of errors. 

See for example activity stream: https://github.com/mozilla/activity-stream-l10n/blob/master/l10n/templates/strings.properties#L35

locales/en-US/snoozetabs.properties
```diff
@@ -0,0 +1,45 @@
...
+timeThisWeekend=This Weekend
+timeNextWeek=Next Week
+timeNextMonth=Next Month
+# LOCALIZATION NOTE: This will be the next time Firefox opens.
+timeNextOpen=Next Open
+timeNextOpenLong=the next time Firefox opens
+timePickADate=Pick a Date/Time
+timeUpcomingToday=later today at $TIME$
+timeUpcomingTomorrow=tomorrow at $TIME$
+timeUpcomingLater=$WEEKDAY$, $MONTH$ $DATE$ at $TIME$
```
flodolo: What is $DATE$ here? Is it the number of the day?
How do you format $TIME$ (separator, 24h vs 12h)?
Can localizers remove $WEEKDAY$ for example?

We definitely need to include that info in the comment, and maybe give a practical example.

locales/en-US/snoozetabs.properties
bwinton: Changed!  🙂

locales/en-US/snoozetabs.properties
bwinton: Fixed!

locales/en-US/snoozetabs.properties
bwinton: Added!

locales/en-US/snoozetabs.properties
bwinton: Changed!

Uh, people can't remove stuff yet, but I think I'll need to add that.

pontoon-to-webext.js
```diff
@@ -0,0 +1,134 @@
...
+  for (const entry of entries) {
+    const publicPath = path.join(process.cwd(), config.dest, entry.locale.replace('-', '_'));
+    const localesPath = path.join(publicPath, 'messages.json');
+
+    FS.makeTree(publicPath).then(() => {
+      FS.write(localesPath, JSON.stringify(entry.content, null, 2))
+      .then(() => {
+        log(`Done compiling locales at: ${localesPath}`);
+      }).catch((e) => {
+        log(e);
```
pdehaan: Do we want `log(e)` or `error(e)`?

pontoon-to-webext.js
```diff
@@ -0,0 +1,134 @@
...
+      });
+    }).catch((e) => {
+      log(e);
```
pdehaan: Do we want `log(e)` or `error(e)`?

pontoon-to-webext.js
```diff
@@ -0,0 +1,134 @@
...
+    });
+  }
+}
+
+function readPropertiesFile(filePath) {
+  return new Promise((resolve, reject) => { // eslint-disable-line promise/avoid-new
```
pdehaan: Should we just disable the `promise/avoid-new` ESLint rule?
It's totally legit Promise usage.

pontoon-to-webext.js
```diff
@@ -0,0 +1,134 @@
...
+      });
+    }).catch((e) => {
+      reject(e);
+    });
+  });
+}
+
+function processMessageFiles(locales) {
+  if (!locales) {
+    error('List of locales was undefined. Cannot run pontoon-to-json.');
```
pdehaan: Should these be "pontoon-to-json", or "pontoon-to-webext"?

test/lib/metrics-test.js
```diff
@@ -3,7 +3,8 @@ const packageMeta = require('../../package.json');
 import { expect } from 'chai';
 import sinon from 'sinon';
 
-import { PICK_TIME } from '../../src/lib/times';
+// import { PICK_TIME } from '../../src/lib/times';
+const PICK_TIME = 'pick';
```
pdehaan: What was the reason for hardcoding the constant here?

pontoon-to-webext.js
bwinton: Fixed.

pontoon-to-webext.js
bwinton: Fixed!

pontoon-to-webext.js
bwinton: Well, I would say 'nah'…  We don't use it _that_ much and only in this file…

test/lib/metrics-test.js
bwinton: Test failures importing `times.js` without it.  😢

locales/en-US/snoozetabs.properties
bwinton: Added!

pontoon-to-webext.js
pdehaan: Sure, sounds good.
If you want, you should be able to disable the rule per-file by just adding this comment at the top (and not needing to sprinkle it throughout):

```js
/* eslint-disable promise/avoid-new */
```

pontoon-to-webext.js
bwinton: Fixed.

src/background.js
```diff
@@ -6,15 +6,21 @@
 
 'use strict';
 
+import { makeLogger } from './lib/utils';
+const log = makeLogger('BE');
+
 import moment from 'moment';
+import 'moment/min/locales.min';
+browser.i18n.getAcceptLanguages().then(languages => {
```
lmorchard: Could this `getAcceptLanguages()` dance get done once in `init()` up in `src/popup/snooze-popup.js`, rather than in every component? Then, also pass in a ref to `getMessage()` alongside other functions like `scheduleSnoozedTab()`?

bwinton: Sure, I'll give that a try!

bwinton: And fixed.  Way cleaner this way, too!  😃

src/lib/components/ManagePanel.js
```diff
@@ -12,12 +11,12 @@ export default class ManagePanel extends React.Component {
     this.state = {
       datepickerActive: false,
       editedItem: null,
-      dateChoice: moment()
+      dateChoice: this.props.moment()
```
lmorchard: Could just use `props` here, since it's passed into the constructor. But, guessing this was a find/replace maybe

pontoon-to-webext.js
```diff
@@ -0,0 +1,164 @@
+#! /usr/bin/env node
+
+/* eslint-disable promise/avoid-new */
+
+const propertiesParser = require('properties-parser');
+const path = require('path');
+const FS = require('q-io/fs');
+const argv = require('minimist')(process.argv.slice(2));
+
+const Habitat = require('habitat');
```
lmorchard: Where / how is this getting used? I don't see anything other than `Habitat.load()` around here

lmorchard:
For the missing `browser` APIs in tests, [you can define `sinon.spy` stubs on `global.browser`](https://github.com/bwinton/SnoozeTabs/blob/f56b885f31619fa5af486784be62caf68882d5bf/test/lib/metrics-test.js#L23) in the tests. But, since this stuff is called on module import, we might need to [bring in a `test/.setup.js`](https://github.com/lmorchard/SnoozeTabs/commit/b193226#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R76) to populate `global.browser` before any of the imports happen.

src/lib/times.js
```diff
@@ -1,22 +1,32 @@
...
+  {id: 'weekend', icon: 'weekends.svg', title: browser.i18n.getMessage('timeThisWeekend')},
+  {id: 'week', icon: 'next_week.svg', title: browser.i18n.getMessage('timeNextWeek')},
+  {id: 'month', icon: 'next_month.svg', title: browser.i18n.getMessage('timeNextMonth')},
+  {id: NEXT_OPEN, icon: 'next_open.svg', title: browser.i18n.getMessage('timeNextOpen')},
+  {id: PICK_TIME, icon: 'pick_date.svg', title: browser.i18n.getMessage('timePickADate')},
 ];
 
 if (process.env.NODE_ENV === 'development') {
-  times.unshift({id: 'debug', icon: 'nightly.svg', title: 'Real Soon Now!'});
+  times.unshift({ id: 'debug', icon: 'nightly.svg', title: browser.i18n.getMessage('timeLaterToday')});
```
lmorchard: `timeLaterToday` here should be `timeRealSoonNow`

src/lib/components/ManagePanel.js
bwinton: Doh!  Fixed.

package.json
```diff
@@ -63,6 +67,7 @@
     "run": "web-ext run -s dist --firefox=nightly",
     "clean": "rm -rf dist && shx mkdir -p dist/popup",
     "watch": "npm-run-all --parallel watch:*",
+    "copy:locales": "node pontoon-to-webext",
```
lmorchard: Probably worth adding a `watch:locales` around here, too, so that this `copy:locales` gets run whenever the locales are updated.

pontoon-to-webext.js
bwinton: My understanding of Habitat is that it loads stuff from a `.env` file into `process.env`, so it's not exactly used directly… Having said that, the code comes from `pontoon-to-json`, so I'ld kind of like to leave it in, if that's okay.  😉

src/lib/times.js
bwinton: Fixed.  (I fixed that before, too, but must have lost it…)

pontoon-to-webext.js
lmorchard: Eh, okay, thought maybe it was a leftover from other tinkering that got deleted. Because I do that all the time. :man_shrugging:

package.json
bwinton: Done.

lmorchard:
Oh, and we could leave the hack in there for now to make tests pass, and follow up with an improvement later to add `test/.setup.js`

pontoon-to-webext.js
pdehaan: :+1: I think we should add a little note in the README.md or whatever about creating an .env file and adding the `SUPPORTED_LOCALES` config (or whatever ENV variables we're trying to import/load via **habitat**).

PR #173, opened 2017-02-14
Improve localization

flodolo:
I kept separate commits on purpose.

Commit 1:
* Rename timeNextOpenLong to timeUpcomingNextOpen. As it is, it's completely unclear where it's used (I assumed as a tooltip somewhere)
* Change placeholder in confirmTimeTitle to make it clear it’s not an actual time
* Reorganize strings more logically
* Improve localization comments

Commit 2:
Split a string in two just to display it on two lines is quite awful, and will probably lead to poor translations if localizers don't test. I'm pretty sure those 4 lines of JS are cringeworthy, but at least you avoid injecting raw HTML in the first span.

Commit 3:
Italian localization, since I've updated the original file (currently in the l10n branch) to test. You will have to ignore the one in the branch when merging.

### DIFF
--- a/locales/en-US/snoozetabs.properties
+++ b/locales/en-US/snoozetabs.properties
@@ -13,8 +13,8 @@ confirmIconAltText=tab icon
 # https://github.com/bwinton/SnoozeTabs/issues/60
 # {time_exp} will be replaced by one of the timeUpcoming* strings, e.g.
 # "later today at 3.00pm", "the next time Firefox opens", etc.
-# You might need to translate this as 'This tab will snooze until: {timeexp}.'
-confirmTimeTitle=This tab will snooze until {timeexp}.
+# You might need to translate this as 'This tab will snooze until: {time_exp}.'
+confirmTimeTitle=This tab will snooze until {time_exp}.
 timeUpcomingNextOpen=the next time Firefox opens
 # "{time}" is the time the tab will snooze until.
 timeUpcomingToday=later today at {time}
--- a/locales/en-US/snoozetabs.properties
+++ b/locales/en-US/snoozetabs.properties
@@ -1,35 +1,45 @@
-# LOCALIZATION NOTE: This add-on will let you select tabs to close for a while, after which they will re-appear in your browser with a notification.
+# This add-on will let you select tabs to close for a while, after which they
+# will re-appear in your browser with a notification.
 extDesc=An add-on to let you snooze your tabs for a while.
 popupTitle=Snooze tab until…
 contextMenuTitle=Snooze tab until…
 bookmarkFolderTitle=Snoozed Tabs
 notificationTitle=Tab woke up…
 
+# Alternative text for the tab icon
 confirmIconAltText=tab icon
-# LOCALIZATION NOTE: "{time}" is the time the tab will snooze until, as per https://github.com/bwinton/SnoozeTabs/issues/60
-confirmTimeTitle=This tab will snooze until {time}.
+
+# "{time_exp}" is the time the tab will snooze until, see also
+# https://github.com/bwinton/SnoozeTabs/issues/60
+# {time_exp} will be replaced by one of the timeUpcoming* strings, e.g.
+# "later today at 3.00pm", "the next time Firefox opens", etc.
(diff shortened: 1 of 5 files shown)


src/lib/components/ManagePanel.js
```diff
@@ -86,7 +86,8 @@ export default class ManagePanel extends React.Component {
 
   getDate(time) {
     if (time === NEXT_OPEN) {
-      return browser.i18n.getMessage('manageDateNext');
+      const message = browser.i18n.getMessage('manageDateNextOpen');
+      return message.split('<br>', '2')[0] || message;
```
bwinton: Ewww!  🙂

locales/en-US/snoozetabs.properties
```diff
@@ -1,35 +1,45 @@
-# LOCALIZATION NOTE: This add-on will let you select tabs to close for a while, after which they will re-appear in your browser with a notification.
+# This add-on will let you select tabs to close for a while, after which they
```
bwinton: Hmm, do we not need `LOCALIZATION NOTE:` at the start of comments to have them show up in Pontoon?

locales/en-US/snoozetabs.properties
```diff
@@ -1,35 +1,45 @@
...
 contextMenuTitle=Snooze tab until…
 bookmarkFolderTitle=Snoozed Tabs
 notificationTitle=Tab woke up…
 
+# Alternative text for the tab icon
 confirmIconAltText=tab icon
-# LOCALIZATION NOTE: "{time}" is the time the tab will snooze until, as per https://github.com/bwinton/SnoozeTabs/issues/60
-confirmTimeTitle=This tab will snooze until {time}.
+
+# "{time_exp}" is the time the tab will snooze until, see also
```
bwinton: `{time_exp}` or `{timeexp}`?

locales/en-US/snoozetabs.properties
flodolo: Historically we used `LOCALIZATION NOTE (string_ID):`, the reality is that Pontoon shows the comment before the string, there's no actual way to bind the comment to a string (or a group of strings).

locales/en-US/snoozetabs.properties
flodolo: `{time_exp}` doesn't work out of the box. I haven't debugged why, but I assume because of [this regex](https://l10n.mozilla-community.org/webstatus/?product=activity-stream). Can look into it if needed.

src/lib/components/ManagePanel.js
flodolo: Yeah… At least it "breaks" gracefully, and it seems less scary than injecting raw HTML (it's from a safe-ish source, but still…). And, if you change your mind about that, you can still use the same string.

locales/en-US/snoozetabs.properties
bwinton: Wrong link, but if you wanted to fix it, I think you could change [this line](https://github.com/bwinton/SnoozeTabs/blob/master/pontoon-to-webext.js#L97) and [this line](https://github.com/bwinton/SnoozeTabs/blob/master/pontoon-to-webext.js#L128), and have it all work.
(Maybe pull that regex out into a constant, and make it `/\{([A-Za-z0-9_@]*)\}/g` as per [the docs](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/i18n/Locale-Specific_Message_reference#name)?)

flodolo: Damned clipboard… That was the correct URL, but I only spotted one. Let me look into it.

flodolo:
@bwinton Updated if you want to double check (not really a dev here :-))

bwinton:
The changes look good to me.  I trust you tested it and it seems to work?

flodolo:
Yep, this is running on my branch

schermata 2017-02-14 alle 18 11 25

schermata 2017-02-14 alle 18 11 35

flodolo:
Not sure how this is going to end up in the l10n branch, but I trust you with that :-)

Once this landed, I need a couple of info in the other issues and I'm ready to enable the project in Pontoon

PR #175, opened 2017-02-14
fix: Remove the panel shadow when in narrow menu panel

### LINKED ISSUE #142: Also, when button is moved in the Firefox menu, the gray margin does not cover the entire panel.

lmorchard:
Fixes #142.

### DIFF
--- a/src/popup/snooze.scss
+++ b/src/popup/snooze.scss
@@ -589,16 +589,17 @@ li.entry {
 
   .panel {
     height: 100%;
-    left: calc(100% + 7px);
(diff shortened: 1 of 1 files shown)


lmorchard:
Rather than try to get the shadow to apply to the full panel height, we talked about just removing it in last week's status meeting

Issue #110, opened 2017-01-19
Generate manifest.json using package.json data on build?


lmorchard:
So, with SDK addons, we stick the metadata in `package.json`. For webext, we stick metadata in `manifest.json`. Stubbed my brain on this difference while working on #102 and trying to bump version numbers. 

Since we have a build process, it might be handy to have `manifest.json` be a templated & generated artifact based on data from `package.json`.

pdehaan:
@lmorchard What fields in manifest.json would be variable?
- `version` &mdash; definitely.
- `name` &mdash; maybe, but how likely are we to rename this?
- `description` &mdash; maybe?
- `author` &mdash; package.json is Blake, src/manifest.json is Mozilla
- `homepage` => `homepage_url` &mdash; maybe?

lmorchard:
Version's definitely variable - and tracking down errors in the build process is what led me to this.

As for the rest, the data is all mostly redundant between two JSON files in two different spots. Also potentially inconsistent (e.g. author). Thinking this will all also be useful example stuff for the next Test Pilot experiment.

pdehaan:
Here's a super weak attempt (./bin/sync-manifest.js) which keeps the `version` in sync, but should be pretty trivial to extend to other properties:

```sh
#!/usr/bin/env node

const { writeFileSync } = require('fs');

const manifestPath = require.resolve('../src/manifest.json');

const { version } = require('../package.json');
const manifest = require(manifestPath);

const data = Object.assign(manifest, { version });

writeFileSync(manifestPath, JSON.stringify(data, null, 2) + '\n');
```

I imagine it may be easiest to add this to a npm `prebuild` script hook, so the manifest.json gets automatically updated before `npm run build` gets called.

bwinton:
We would also want to run it on watch, so that I use the fancy "Load Temporary Add-on" button…  🙂
In another project, I used [a gulp thing](https://github.com/bwinton/whimsy/blob/WebExtensionAPI/gulpfile.js#L28) to do it, and that seemed to work out pretty well!

PR #184, opened 2017-02-14
Bump to v1.0.13

lmorchard:
Just pushed master to production at v1.0.12. So, this PR gets us ready for the next push

### DIFF
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -14,7 +14,7 @@
   "icons": {
     "48": "icons/bell_icon.svg"
   },
-  "version": "1.0.12",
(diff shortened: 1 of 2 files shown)


src/manifest.json
```diff
@@ -14,7 +14,7 @@
   "icons": {
     "48": "icons/bell_icon.svg"
   },
-  "version": "1.0.12",
+  "version": "1.0.13",
```
bwinton: It's a shame we don't auto-generate one of these from the other…  😉

lmorchard: Yeah, that's #110

PR #188, opened 2017-02-17
fix: Add `update_url` to enable Test Pilot updates.

### LINKED ISSUE #187: Add an updateURL
Please add an updateURL.  I think you should just need to add this right after the `id` line at https://github.com/bwinton/SnoozeTabs/blob/master/src/manifest.json#L4 :

"update_url": "https://testpilot.firefox.com/files/snoozetabs@mozilla/updates.json"

bwinton:
Fixes #187.

### DIFF
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -1,7 +1,8 @@
 {
   "applications": {
     "gecko": {
-      "id": "snoozetabs@mozilla.com"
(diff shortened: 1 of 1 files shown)


pdehaan:
👍  Per https://developer.mozilla.org/en-US/Add-ons/Updates

PR #189, opened 2017-02-17
fix: Collapse empty space at bottom of main panel in prod build

lmorchard:
Noticed in the production build, without the "Real Soon Now" option, that there was empty space at the bottom of the panel. Tried taking out the fixed height to let it size itself, and it seemed to work. 

I think we needed the absolute height at one point for whatever shenanigans I was up to with absolute positioning in the CSS and content determining the size of the panel, but that seems to no longer be the case.


lmorchard:
Oh, wait, this borks the date panels. Back to the drawing board.

PR #190, opened 2017-02-17
fix: Collapse empty space at bottom of main panel in prod build

lmorchard:
Maybe let's try this again? This lets the panel resize to the number of items in the main panel. So when "Real Soon Now" is gone in production builds, there's not an empty space at the bottom.

I also tweaked the calendar date cell size to shrink it a little.

### DIFF
--- a/src/popup/snooze.scss
+++ b/src/popup/snooze.scss
@@ -2,7 +2,6 @@
 @import 'node_modules/rc-time-picker/assets/index';
 
 $panel-width: 320px;
-$panel-height: 474px;
 $panel-shadow-width: 13px;
 
 @font-face {
@@ -34,7 +33,6 @@ body {
 
 .panel-wrapper {
   background-color: #333333;
-  height: $panel-height;
 
   >div {
     height: 100%;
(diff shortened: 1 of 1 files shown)


pdehaan:
@lmorchard I think this needs a rebase.

pdehaan:
LGTM using `$ NODE_ENV=production npm start` on macOS Sierra:

washington_post__breaking_news__world__us__dc_news___analysis_-_the_washington_post_and_nightly

(versus the current behavior from the version hosted on Test Pilot dev)

menubar_and_firefox_test_pilot_-_snooze_tabs_and_browser_console_and_fix__collapse_empty_space_at_bottom_of_main_panel_in_prod_build_by_lmorchard_ _pull_request__190_ _bwinton_snoozetabs

bwinton:
Uh, looks a little weird with only one snoozed tab, and forced into manage mode (by launching from `about:debugging` for example).

bwinton:
Maybe we could have a `min-height` of `350px` or something?

lmorchard:
Hmm, seems to be more of a can of worms here than just min-height fixes. Will keep poking, but may tag a prod release before figuring this out

lmorchard:
Going to close this, rather than let it kind of hang out in a PR. Needs more jiggerypokery

PR #193, opened 2017-02-21
fix: Correctly localize dates and times.

### LINKED ISSUE #172: [l10n] Dates and time are not correctly localized
* Time is set to 12h, it should be 24h for Italian
* Abbreviated day name is uppercase, Abbreviated month name lowercase. Not the end of the world, but it looks kind of strange
* Order is wrong, day number should be before the month name

schermata 2017-02-14 alle 12 15 51

bwinton:
Add some unfortunately-less-localizable date formats, and use those instead.
(I rather liked the IIFE to populate the formats based on which language you were using at install-time. 😉)

Fixes #172.

### DIFF
--- a/src/lib/times.js
+++ b/src/lib/times.js
@@ -2,6 +2,7 @@
 
 import moment from 'moment';
 import 'moment/min/locales.min';
+import {formats} from './time-formats';
 moment.locale(browser.i18n.getUILanguage());
 
 const NEXT_OPEN = 'next';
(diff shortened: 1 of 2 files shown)


src/lib/times.js
```diff
@@ -22,33 +22,66 @@ if (process.env.NODE_ENV === 'development') {
   times.unshift({ id: 'debug', icon: 'nightly.svg', title: browser.i18n.getMessage('timeRealSoonNow')});
 }
 
+
+const i18n_formats = ((locale) => {
+  const formats = {
+    default: {
```
lmorchard: Might be worth splitting these out into a separate JSON file, [which we can import like a JS module](https://github.com/bwinton/SnoozeTabs/blob/master/src/lib/metrics.js#L7). Thinking that could make it easier for localizers to find and add entries.

src/lib/times.js
```diff
@@ -2,6 +2,7 @@
 
 import moment from 'moment';
 import 'moment/min/locales.min';
+import {formats} from './time-formats';
```
lmorchard: :+1:

PR #200, opened 2017-02-22
chore: Pre-bump version to 1.0.15

lmorchard:
Just released v1.0.14, so we're working toward 1.0.15 now

### DIFF
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -15,7 +15,7 @@
   "icons": {
     "48": "icons/bell_icon.svg"
   },
-  "version": "1.0.14",
+  "version": "1.0.15",
 
   "permissions": [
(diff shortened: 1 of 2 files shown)


src/manifest.json
```diff
@@ -15,7 +15,7 @@
   "icons": {
     "48": "icons/bell_icon.svg"
   },
-  "version": "1.0.14",
+  "version": "1.0.15",
```
bwinton: Geez, we should really fix this duplication sometime…
Maybe I'll con @ericawright into doing it when I'm out of town…  😉

lmorchard: Yeah, I've just been too lazy / busy to work on it 😞

PR #205, opened 2017-02-23
(WIP) Add custom datetime formats for more locales, fix manage panel

flodolo:
Need a lot of feedback on the localized time formats


flodolo:
After chatting with zibi over IRC, I want to check if we can rely on CLDR and Intl to simplify things easier.

flodolo:
Let me close this one, will continue the discussion in the issue. Too much wrong going on here.

PR #206, opened 2017-02-23
fix: Use Intl for formatting dates and times.

### LINKED ISSUE #204: Use Intl for formatting date and times in toolbar panel
Right now we have custom formats only for Italian and German, we should cover all locales working on it.

bwinton:
Also remove moment-locales to save space.

Fixes #204.

### DIFF
--- a/locales/en-US/snoozetabs.properties
+++ b/locales/en-US/snoozetabs.properties
@@ -20,9 +20,9 @@ timeUpcomingNextOpen=the next time Firefox opens
 timeUpcomingToday=later today at {time}
 # "{time}" is the time the tab will snooze until.
 timeUpcomingTomorrow=tomorrow at {time}
-# "{weekday}" is the day of the week, "{day}" is the day of the month.
+# "{date}" is the full date of the month.
 # Example: "Sun, Mar 5 at 7:05am".
-timeUpcomingLater={weekday}, {month} {day} at {time}
+timeUpcomingOther={date} at {time}
 
 confirmOkButton=OK
 confirmCancelButton=Cancel
--- a/src/lib/time-formats.js
+++ b/src/lib/time-formats.js
@@ -1,14 +1,50 @@
-/* exported formats */
+/* exported getLocalizedDateTime */
 
-export const formats = {
-  default: {
-    short_time: '[@] ha',
-    short_date_time: 'ddd [@] ha',
-    long_date_time: 'ddd MMM D \ [@] ha'
-  },
-  'it': {
-    short_time: '[@] H',
-    short_date_time: 'ddd [@] H',
-    long_date_time: 'ddd D MMM \ [@] H'
+const uiLocales = [browser.i18n.getUILanguage().replace('_', '-'), 'en-US'];
+
+const formats = {
+  'date_day': new Intl.DateTimeFormat(uiLocales, { weekday: 'short', month: 'short', day: 'numeric' }),
+  'date_year': new Intl.DateTimeFormat(uiLocales, { month: 'short', day: 'numeric', year: 'numeric' }),
+  'long_date_time': new Intl.DateTimeFormat(uiLocales, { weekday: 'short', month: 'short', day: 'numeric', hour: 'numeric', minute: 'numeric' }),
+  'short_date_time': new Intl.DateTimeFormat(uiLocales, { weekday: 'short', hour: 'numeric', minute: 'numeric' }),
+  'short_time': new Intl.DateTimeFormat(uiLocales, { hour: 'numeric', minute: 'numeric' }),
+
+  'confirmation_time': new Intl.DateTimeFormat(uiLocales, { hour: 'numeric', minute: 'numeric' }),
+  'confirmation_time_no_minutes': new Intl.DateTimeFormat(uiLocales, { hour: 'numeric', minute: 'numeric' }),
+  'confirmation_time_no_minutes-en': new Intl.DateTimeFormat(uiLocales, { hour: 'numeric' }),
+  'confirmation_date': new Intl.DateTimeFormat(uiLocales, { weekday: 'short', month: 'short', day: 'numeric' }),
+
+  'long_date_time-en': [
+    new Intl.DateTimeFormat(uiLocales, { weekday: 'short', month: 'short', day: 'numeric' }),
+    new Intl.DateTimeFormat(uiLocales, { hour: 'numeric' })
+  ],
+  'short_date_time-en': [
+    new Intl.DateTimeFormat(uiLocales, { weekday: 'short' }),
+    new Intl.DateTimeFormat(uiLocales, { hour: 'numeric' })
+  ],
+  'short_time-en': [
+    {format: () => ''},
+    new Intl.DateTimeFormat(uiLocales, { hour: 'numeric' })
+  ],
+};
+
+export const getLocalizedDateTime = function (time, format) {
+  let formattedDateTime;
+
+  let formatter = formats[format];
+  if (formatter) {
+    const date = time.toDate();
+    const realLocale = formatter.resolvedOptions().locale;
+    if (realLocale.startsWith('en') && formats[format + '-en']) {
+      formatter = formats[format + '-en'];
+      if (formatter[0]) {
+        formattedDateTime = formatter[0].format(date) + ' @ ' + formatter[1].format(date);
+      }
+    }
+    if (!formattedDateTime) {
+      formattedDateTime = formatter.format(date);
+    }
   }
+
+  return formattedDateTime;
 };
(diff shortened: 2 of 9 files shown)


src/lib/time-formats.js
```diff
@@ -1,14 +1,34 @@
...
-    short_time: '[@] ha',
-    short_date_time: 'ddd [@] ha',
-    long_date_time: 'ddd MMM D \ [@] ha'
-  },
-  'it': {
-    short_time: '[@] H',
-    short_date_time: 'ddd [@] H',
-    long_date_time: 'ddd D MMM \ [@] H'
+export const getLocalizedDateTime = function (time, format) {
+  const uiLocale = browser.i18n.getUILanguage().replace('_', '-') || 'en-US';
```
zbraniecki: you don't need to replace '_' for '-'. Intl API handles both properly.

src/lib/time-formats.js
```diff
@@ -1,14 +1,34 @@
...
+  let formattedDateTime;
+
+  switch (format) {
+    case 'date_day':
+      formattedDateTime = time.toDate().toLocaleString(uiLocale, {weekday: 'short', month: 'short', day: 'numeric'});
```
zbraniecki: it seems like you could extract the date from the object once at the top of the function instead of in each individual call?

src/lib/time-formats.js
flodolo: It didn't work for me (and took a bit to figure out why the add-on wasn't working)
```
(new Date()).toLocaleDateString('zh_TW', {hour: 'numeric'});
/*
Exception: RangeError: invalid language tag: zh_TW
@Scratchpad/3:10:2
*/
````

Comment (zbraniecki) [inline src/lib/time-formats.js]: Oh, good point. It seems that ICU is way more promiscuous than Intl API is. Weird as I'm pretty sure we're trying to stay lenient on what we accept.

File: src/lib/time-formats.js (hunk 2 above)
Comment (bwinton) [inline src/lib/time-formats.js]: Done.

File: src/lib/time-formats.js (hunk 1 above)
Comment (bwinton) [inline src/lib/time-formats.js]: Done, then undone.  😉

**Comment by bwinton** [comment]
> Instead of calling to Date.prototype.toLocaleString, you can create a Intl.DateTimeFormat objects, and store them. The costly part of formatting is then done once, and the format operation is cheap.

I thought about that, but it's kind of nice to not have to reload the add-on when you change your locale…  I think I'm not going to make this change until we find that the extra complexity/lack of flexibility is worth it.

File: src/lib/time-formats.js (hunk 1 above)
Comment (flodolo) [inline src/lib/time-formats.js]: I wonder if there's a good way to fall back to en-US here, to make the code a bit more stable.

Comment (bwinton) [inline src/lib/time-formats.js]: Yeah, probably.  I'll see what I can come up with…

Comment (flodolo) [inline src/lib/time-formats.js]:
```JS
try {
  (new Date()).toLocaleString(uiLocale);
} 
catch (e) {
  // Locale is not supported by ICU
  uiLocale = 'en-US';
}
```

zbraniecki: Just do `let uiLocale = [uiLocale, 'en-US'].

Intl API accept language fallback chain.

flodolo:
```JS
let uiLocale = ['zh_TW', 'en-US'];
(new Date()).toLocaleString(uiLocale);
/*
Exception: RangeError: invalid language tag: zh_TW
@Scratchpad/4:2:2
*/
```

zbraniecki: oh, I mean, canonicalize the uiLocale, but turn it into a fallback list after and add `en-US` as a last option.

flodolo: Uhm, not sure if it would be more robust than the try…catch. 

My doubt is if there are locales that could break the code even with that replacement.

bwinton: Hmm, unless we could tell which locale the DateTimeFormatter ended up using, that wouldn't work out so well with the `uiLocale.startsWith('en')` check down below…

zbraniecki: Oh, you mean for broken locale code, not for missing locale.

I would assume that browser.i18n.getUILocale canonicalize it's output. We should not be able to produce non-canonical locale IDs.

flodolo: I don't think showing `@ ha` for these locales would be a problem, I expect this to be a corner case, possibly that we never hit.

zbraniecki: > Hmm, unless we could tell which locale the DateTimeFormatter ended up using,

If you use `Intl.DateTimeFormat` then you can:

```javascript
let fmt = new Intl.DateTimeFormat(['zh-TW', 'en-US'], options);
fmt.resolvedOptions().locale;
```

bwinton: Done.  Thanks!

bwinton:
Okay, I did a little more refactoring, but I think this is pretty decent.  Screenshots of other languages would be appreciated, as would comments on more things to fix.  😉

flodolo:
> Okay, I did a little more refactoring, but I think this is pretty decent.

Is it just me or I don't see changes in the code? Missed push?

As for testing with other locales, I discovered it's relatively easy: 
* Copy locales from l10n branch
* Snooze a bunch of tabs while the add-on is in a language you understand
* Change general.useragent.locale in about:config to whatever locale you want to test
* Save a file to retrigger the add-on build

Fun extra: you get to break the add-on sometimes, with a double panel vertically stacked 🎆

bwinton:
Dang, I always forget the `git add` step!  The new code should be there now…  🙄

flodolo:
zibi is definitely a better pick to look at the code. 

From testing it looks great as far as I can tell. 

There is a small bug (it was already there in the previous code): when you select a date from next year, it won't show the year in the notification, just day+month+time, which is confusing. The panel has that information on the other hand.

src/lib/time-formats.js
```diff
@@ -1,14 +1,50 @@
...
+  ],
+};
+
+export const getLocalizedDateTime = function (time, format) {
+  let formattedDateTime;
+
+  let formatter = formats[format];
+  if (formatter) {
+    const realLocale = formatter.resolvedOptions().locale;
+    formattedDateTime = formatter.format(time.toDate());
```
flodolo: I wonder if this should be a return after the if instead, since you discard this value for en-* locales

bwinton:
> There is a small bug (it was already there in the previous code): when you select a date from next year, it won't show the year in the notification, just day+month+time, which is confusing. The panel has that information on the other hand.

Yeah, can you file that as a new bug, and ask Sevaan what he thinks?  I know how we could do it, and it's not that complicated, but I'ld like to get his opinion on when we should show the year, and when not to.

src/lib/time-formats.js
bwinton: I kinda thought that, but the format is apparently pretty cheap, and so I don't mind throwing away this value in some cases.  :smile:

bwinton: Actually, I need to refactor it to pull out the `time.toDate()` anyways, so let's fix that in a different way.

flodolo:
Found another bug: by dropping moment localization, you lose localization for the datetime picker :-\

bwinton:
(I'm terrible at this.  I _think_ that's the last thing I needed to push.)

bwinton:
> Found another bug: by dropping moment localization, you lose localization for the datetime picker :-\

Dammit.  🙁

locales/en-US/snoozetabs.properties
```diff
@@ -20,9 +20,9 @@ timeUpcomingNextOpen=the next time Firefox opens
 timeUpcomingToday=later today at {time}
 # "{time}" is the time the tab will snooze until.
 timeUpcomingTomorrow=tomorrow at {time}
-# "{weekday}" is the day of the week, "{day}" is the day of the month.
+# "{date}" is the full date of the month.
 # Example: "Sun, Mar 5 at 7:05am".
-timeUpcomingLater={weekday}, {month} {day} at {time}
+timeUpcomingOther={date} at {time}
```
zbraniecki: This is great to see as this pattern corresponds 1-1 to a pattern in CLDR that we currently don't expose, but hopefully will one day :)
[0] http://www.unicode.org/cldr/charts/30/summary/en.html#1945

src/lib/time-formats.js
```diff
@@ -1,14 +1,50 @@
...
-    short_date_time: 'ddd [@] ha',
-    long_date_time: 'ddd MMM D \ [@] ha'
-  },
-  'it': {
-    short_time: '[@] H',
-    short_date_time: 'ddd [@] H',
-    long_date_time: 'ddd D MMM \ [@] H'
+/* exported getLocalizedDateTime */
+
+const uiLocale = [browser.i18n.getUILanguage().replace('_', '-'), 'en-US'];
```
zbraniecki: `uiLocales` maybe to indicate that it's a language fallback chain, not a single locale?

locales/en-US/snoozetabs.properties
bwinton: 🙌 🙂

src/lib/time-formats.js
bwinton: (fixed.)

PR #209, opened 2017-02-23
fix: snooze_time should be a GA metric, not dimension

lmorchard:
Still kind of stumbling toward understanding GA, and it seems like the integer `snooze_time` should be handled as a metric to measure and not a dimension by which to group measurements.


lmorchard:
> (Feel free to get another review from someone who knows more metrics stuff.)

Kinda feel like we're all kind of learning this stuff %)

PR #215, opened 2017-02-24
Display timepicker as 12h/24h depending on locale

flodolo:


flodolo:
Thanks @zbraniecki  for the code tip on detecting 12h vs 24h

## Italian
schermata 2017-02-24 alle 19 13 00

## English
schermata 2017-02-24 alle 19 14 26

lmorchard:
Hmm, this changes the display format in the field, but not the controls in the time picker itself. It's still a 24hr widget. I haven't yet figured out if that's an easy switch in the widget, or if we have to do something else entirely

flodolo:
> Hmm, this changes the display format in the field, but not the controls in the time picker itself. It's still a 24hr widget. I haven't yet figured out if that's an easy switch in the widget, or if we have to do something else entirely

Correct, it's only affecting the way the time is displayed (which is wrong for most languages).

flodolo:
Thanks for reviewing, I suppose I can merge this (if I shouldn't, I can always blame @bwinton for giving me the keys 🔑 )

src/lib/components/DatePickerPanel.js
```diff
@@ -36,6 +36,12 @@ export default class DatePickerPanel extends React.Component {
     const { currentValue, confirmDisabled } = this.state;
     const disabledTimeFns = this.disabledTime();
 
+    const uiLocale = [browser.i18n.getUILanguage().replace('_', '-'), 'en-US'];
+    const dtf = new Intl.DateTimeFormat(uiLocale, {hour: 'numeric'});
+    const timeFormat = (dtf.resolvedOptions().hour12)
+      ? 'h:mm a'
+      : 'HH:MM';
+
```
bwinton: I wonder if we could/should pull this into the new `time-formats.js` file in a future patch?

zbraniecki: I believe you should. Not only because it'll do the hourCycle magic all by itself, but also because not all locales are left-to-right with western arabic numerals ;)
And Intl API handles that for you.

PR #217, opened 2017-02-25
Fix minutes display in timepicker

flodolo:
This is embarrassing, I've used MM (month) instead of mm (minutes) in the format.

This also moves the logic to time-formats.js

### DIFF
--- a/src/lib/components/DatePickerPanel.js
+++ b/src/lib/components/DatePickerPanel.js
@@ -5,6 +5,7 @@ import classnames from 'classnames';
 import 'moment/min/locales.min';
 import Calendar from 'rc-calendar';
 import TimePicker from 'rc-time-picker';
+import { use12hFormat } from '../time-formats.js';
 
 // Arbitrary 0.5s interval for live validation of time selection
 const VALIDATION_INTERVAL = 500;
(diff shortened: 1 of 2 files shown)


src/lib/components/DatePickerPanel.js
```diff
@@ -37,11 +38,7 @@ export default class DatePickerPanel extends React.Component {
     const { currentValue, confirmDisabled } = this.state;
     const disabledTimeFns = this.disabledTime();
 
-    const uiLocale = [browser.i18n.getUILanguage().replace('_', '-'), 'en-US'];
-    const dtf = new Intl.DateTimeFormat(uiLocale, {hour: 'numeric'});
-    const timeFormat = (dtf.resolvedOptions().hour12)
-      ? 'h:mm a'
-      : 'HH:MM';
+    const timeFormat = use12hFormat ? 'h:mm a' : 'HH:mm';
```
bwinton: I think we could move the `?:` part into time-formats, too, and just export a `timeFormat` constant from that file into this file…

flodolo: My reasoning for keeping this part here is that we stopped using time formats all together in that file, since we're using Intl.DateTimeFormat, so this kind of time formatting belongs only to time picker.

I've also checked the component, and I'm afraid you will need to find a replacement if you want to fix the picker only showing 24h
https://github.com/react-component/time-picker/issues/27

bwinton: That makes sense…

PR #235, opened 2017-03-03
refactor: Generate dist/manifest.json from data in package.json

### LINKED ISSUE #110: Generate manifest.json using package.json data on build?

lmorchard:
- bin/generate-manifest.js to create dist/manifest.json based on
  package.json

- Remove original src/manifest.json

- Move pontoon-to-webext.js into bin/

- NPM script changes to use new scripts

Fixes #110.


bin/generate-manifest.js
```diff
@@ -0,0 +1,53 @@
+#! /usr/bin/env node
+
+const path = require('path');
+const fs = require('fs');
+const packageMeta = require('../package.json');
+
+const manifest = {
```
bwinton: Hmm.  I worry about not remembering to edit (or not being able to find) this file when we need to update the permissions or change the icons or something.  What do you think about putting a `manifest.in` somewhere and loading that, instead of "hard-coding" it here.

package.json
```diff
@@ -67,19 +68,21 @@
...
     "package": "npm run build && web-ext build -s dist -a .",
     "lint": "npm-run-all --parallel lint:*",
     "lint:js": "eslint --max-warnings=0 .",
     "lint:sass": "sass-lint --max-warnings 0 --verbose --no-exit",
+    "watch:manifest": "onchange -p -v 'bin/generate-manifest.js' -- npm run bundle:manifest",
     "watch:js": "onchange -p -v 'src/**/*.js' -- npm-run-all --parallel bundle:js test:js",
     "watch:test:js": "onchange -p -v 'test/**/*.js' -- npm-run-all --parallel test:js",
     "watch:css": "onchange -p -v 'src/**/*.scss' -- npm run bundle:css",
-    "watch:assets": "onchange -p -v 'src/{manifest.json,icons/*,popup/*.html}' -- npm run copy:assets",
+    "watch:assets": "onchange -p -v 'src/{icons/*,popup/*.html}' -- npm run copy:assets",
```
bwinton: If we had a `manifest.in` we could watch that, but since we don't (yet), should we watch `bin/generate-manifest` for changes to permissions and stuff, too?

bwinton:
(Oh, and I haven't tested it on Windows yet.)

bin/generate-manifest.js
lmorchard: Well, I figured it would be easier to have it all in one place to edit for the things pulled from package.json and the other non-templated properties. Otherwise, it'll be split between a template file and a build script and possibly be confusing that way too

package.json
lmorchard: There is a watch task for `bin/generate-manifest.js` over here: https://github.com/bwinton/SnoozeTabs/pull/235/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R81

bin/generate-manifest.js
lmorchard: FWIW, I was also considering moving *everything* from manifest.json into package.json, since NPM seems pretty agnostic to custom properties in there. Then, this generation script could just scoop everything out of there and transform it as needed

lmorchard:
Just updated the PR to move `src/manifest.json` content into `package.json` and strip the `bin/generate-manifest.js` script down to just essentials. How's that look?

Issue #124, opened 2017-01-26
Doorhanger view changes if you're managing snoozed tabs while tab is automatically unsnoozed


pdehaan:
Found in 57340e7

---

### Steps to reproduce:
1. Snooze a few tabs into the future (an hour or two will do).
1. Snooze a new tab for "Real Soon Now!" which should snooze it for a few seconds.
1. Before the latest tab has managed to unsnooze, open the SnoozeTabs doorhanger and switch to the "Manage Snoozed Tabs" view.
1. Wait.

### Actual results:
Before the tab has been unsnoozed, the visible panel is Manage Snoozed Tabs.
After the tab has been unsnoozed, the visible panel changes.

### Expected results:
The doorhanger panel should not change as the tab is unsnoozed.

pdehaan:
![snooze-tabs-awaken](https://cloud.githubusercontent.com/assets/557895/22315106/a30639d8-e319-11e6-98e2-b5e009b104bd.gif)

lmorchard:
I think this is fixed with https://github.com/bwinton/SnoozeTabs/pull/240

Issue #114, opened 2017-01-20
[Enhancement] "Manage Snoozed Tabs" panel should remain displayed after deleting a snoozed tab from the list


SoftVision-EmilPasca:
[Notes]:
- As a user, I would like to remove multiple tabs from the list without needing to open the list again for each tab that I want out.

[Affected versions]:
- Nightly 53.0a1
- Snooze Tabs 1.0 (built on 1/20/2017)

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- Have a Firefox profile with the latest Snooze Tabs add-on version (1.0-dev, built on 1/20/2017) loaded as a temporary add-on.
- Have a few snoozed tabs in the "Manage Snoozed Tabs" panel.

[Steps to reproduce]:
1. Open the browser with the profile from prerequisite and navigate to any website.
2. Click the "Snooze Tabs" icon from the main browser toolbar.
3. Click on the "Manage Snoozed Tabs" button.
4. Remove a snoozed tab from the "Manage Snoozed Tabs" panel.

[Expected result]:
- The "Manage Snoozed Tabs" panel doesn't slides over after removing a snoozed tab.

[Actual result]:
- The "Manage Snoozed Tabs" panel slides over after removing a snoozed tab.

[Additional Notes]:
- The requested behavior is already available for "about: " pages since there you don't have the option to snooze them, just to see the snoozed ones.
- Attached a screen recording of the issue:
![snoozetabs_delete_item](https://cloud.githubusercontent.com/assets/20083658/22157792/d1a20050-df41-11e6-9923-889a455e9909.gif)

lmorchard:
I think this is fixed with https://github.com/bwinton/SnoozeTabs/pull/240

SoftVision-EmilPasca:
The issue is no longer reproducible on the latest Snooze Tabs version (1.0.15-dev built on 03/07/2017).

PR #240, opened 2017-03-06
refactor: Better state management in root popup component

lmorchard:
- Let React handle state for the root component, rather than
  re-rendering manually from snooze-content.js

- Stop resetting activePanel to main or manage on every storage change

Issue #124.
Issue #114.


lmorchard:
Seemed like manual re-renders and forcibly setting activePanel to main or manage based on ability to snooze the current tab on every data storage update were causing two issues.

Issue #79, opened 2017-01-09
There is a noticeable delay until the Snooze Tabs dropdown appears


SoftVision-CiprianMuresan:
[Affected versions]:
- Nightly 53.0a1
- Snooze Tabs 1.0 (built on 1/9/2016)

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- Have a Firefox profile with the latest Snooze Tabs add-on version (1.0-dev, built on 1/9/2016) loaded as a temporary add-on.

[Steps to reproduce]:
1. Focus the window with the profile from prerequisites.
2. Click the Snooze Tabs button from the browser toolbar.
3. Observe the time it takes for the dropdown to appear.

[Expected results]:
- The dropdown appears instantly.

[Actual results]:
- There is a noticeable delay until the dropdown appears.

[Notes]:
- Here is a screen recording with the issue:
![there is a noticeable delay until the snooze tabs dropdown appears](https://cloud.githubusercontent.com/assets/18304511/21770404/c6d07d3c-d68b-11e6-8b3a-5c2365522656.gif)

bwinton:
So, I dug into this a little, and it looks like it takes between 600-900ms to go through the `snoozetabs-content.js`'s `init()` method, which seems kinda long.  There's more delay before the `init` even gets called, but I don't know how to measure that, much less fix it…

@lmorchard: What do you think about rendering the html in the `background.js`, and sending it to the panel on open?  It doesn't seem to be particularly dependent on which url the tab is showing (aside from some special urls that only get the manage pane)…  Actually, what about pre-rendering both states, and sending the appropriate one when the panel opens?

lmorchard:
Could even just manually code the initial main panel HTML in `src/popup/snooze.html` - React will swoop in after and apply whatever changes to the DOM it needs. Probably easier than trying to get things to render in the background script. Not sure if that will speed up time to when it's actually ready for interaction, though, since it's only after the first render that click handlers are active. But, it might at least *feel* more ready

bwinton:
I think having it appear sooner would be a plus, even if it wasn't live.

lmorchard:
[...]

PR #241, opened 2017-03-06
fix: Incremental improvements on popup load speed

lmorchard:
- Pre-rendered HTML & uglified JS
- Deferred loading of JS via bootstrap script
- Add rollup-analyze script to inspect size of modules making up the
  popup JS - `./bin/rollup-analyze.js src/popup/snooze-content.js`
- Properly watch for HTML changes

Issue #79.


lmorchard:
I don't think this fully fixes #79, but I've gotten a few nods that it improves things. So, let's see if it will help in the next release.

PR #242, opened 2017-03-06
fix: Enable logging on production builds

lmorchard:
Turns out logging would be super handy to have enabled in production builds as users report issues and look in the console for more information.


lmorchard:
Hmm, not sure about slow downs. I'd assume not much? Not sure chatty except for at some key spots, definitely not inside any loops.

PR #248, opened 2017-03-07
fix: Do not display confirmation bar for tabs snoozed via tab context menu

### LINKED ISSUE #220: Tab Context Menu options does not work
Copied from https://github.com/mozilla/pulse/issues/133

>What happened;
>
> 1. I've right clicked on a tab other than my currently active tab.
> 2. Selected Snooze tab until Tomorrow
> 3. Nothing happened.
>
> Tab should've happened
>
> 1. I've right clicked on a tab other than my currently active tab.
> 2. Selected Snooze tab until Tomorrow
> 3. Tab should've been snoozed and removed from tab bar.
>
> Note: This does work if I disable "Ask for confirmation when snoozing tabs" option. I guess the confirmation is displayed in the inactive tab which makes it impossible to respond to without switching to it.

CC @Nidre

lmorchard:
Fixes #220.

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -319,7 +319,7 @@ if (browser.contextMenus.ContextType.TAB) {
     const title = browser.i18n.getMessage('notificationTitle');
     const [time, ] = timeForId(moment(), info.menuItemId);
     handleMessage({
-      'op': 'schedule',
+      'op': tab.active ? 'schedule' : 'confirm',
       'message': {
         'time': time.valueOf(),
         'timeType': info.menuItemId,


ericawright:
This works as described, but for all tabs snoozed from the context menu - including the active tab, is that intended? (I saw in the bug a comment about only tabs other than the active one)

lmorchard:
Hmm, good question - I didn't bother checking for active tab. Looks like we have access to the tab in that click handler, so maybe we *should* display the confirmation bar in the active tab case

lmorchard:
Updated to display confirmation bar on active tab and to just go ahead & schedule on inactive tabs.

PR #250, opened 2017-03-07
fix: Display year if snooze time is not in current year

### LINKED ISSUE #211: When snoozing a tab in a different year, notification bar only shows date and hour
If I snooze a tab a year from now, e.g. Feb 24 2018 @ 5 PM, notification only shows Feb 24 at 5 PM, which is potentially confusing.

On the other hand, the Manage snoozed tabs panel correctly shows the year when a tab is expected to wake up in a different year.

lmorchard:
Fixes #211.

### DIFF
--- a/src/lib/times.js
+++ b/src/lib/times.js
@@ -69,15 +69,24 @@ export function confirmationTime(time, timeType) {
   }
 
   let rv;
+  const thisYear = moment().year();
   const endOfDay = moment().endOf('day');
   const endOfTomorrow = moment().add(1, 'day').endOf('day');
   const upcoming = moment(time);
 
-  let timeStr = getLocalizedDateTime(upcoming, 'confirmation_time_no_minutes');
+  let timeStr;
   if (upcoming.minutes()) {
     timeStr = getLocalizedDateTime(upcoming, 'confirmation_time');
+  } else {
+    timeStr = getLocalizedDateTime(upcoming, 'confirmation_time_no_minutes');
   }
-  const dateStr = getLocalizedDateTime(upcoming, 'confirmation_date');
+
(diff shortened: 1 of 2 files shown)


src/lib/times.js
```diff
@@ -77,7 +78,8 @@ export function confirmationTime(time, timeType) {
   if (upcoming.minutes()) {
     timeStr = getLocalizedDateTime(upcoming, 'confirmation_time');
   }
-  const dateStr = getLocalizedDateTime(upcoming, 'confirmation_date');
+  const dateStr = getLocalizedDateTime(upcoming,
+    (upcoming.year() === thisYear) ? 'confirmation_date' : 'confirmation_date_with_year');
```
bwinton: Could we switch this to the `let`/`if()` like timeStr?  I'm kinda allergic to the ternary `?:` operator…  (I mean, not totally allergic, but it often seems a little too clever for its own good.  :wink:)

bwinton: Oh, my other thought was that we could only show the date if the year wasn't this year, and the month and day was after today.  (i.e. on Dec 31st, a date of Jan 17th isn't going to be assumed to be in this year.)  But that's not really needed in this patch, and I'm not entirely sure it's a good idea.  🙂

ericawright:
(tested, and it works as described btw)

src/lib/times.js
lmorchard: Well, I was trying to avoid re-running `getLocalizedDateTime` since all that changed was the name of the string. But, I can use an if statement instead

lmorchard:
Rebased against master, updated to use if statement instead of ternary, will merge after circleci passes

PR #255, opened 2017-03-08
refactor: Stop attempting to switch styles based on document width

### LINKED ISSUE #179: “Pick a Date/Time” option appearing inconsistently
Reported by Sharon, we'll add a video to this issue.

lmorchard:
- Remove resize event hook for swapping styles

- Remove .popup-narrow styles

- Styling tweaks to make the normal wide panel styles scroll in the menu

- Only render panel components when active, because their presence
  "off-stage" screws up X scrolling in the menu

Fixes #179.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,6 @@
     "svgo": "^0.7.1",
     "web-ext": "^1.8.1"
   },
-  "homepage": "https://github.com/bwinton/SnoozeTabs#readme",
   "keywords": [
     "snooze",
     "tabs",
@@ -67,19 +66,16 @@
   },
   "webextensionManifest": {
     "manifest_version": 2,
-
     "applications": {
       "gecko": {
         "update_url": "https://testpilot.firefox.com/files/snoozetabs@mozilla/updates.json"
       }
     },
-
     "default_locale": "en_US",
     "description": "__MSG_extDesc__",
     "icons": {
       "48": "icons/bell_icon.svg"
     },
-
     "permissions": [
       "alarms",
       "bookmarks",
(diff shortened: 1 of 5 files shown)


lmorchard:
We decided to kind of abandon ship for now on the magical style change between toolbar and menu invocation cases. This PR should clean things up enough so that the menu sidebar is at least usable with a horizontal scrollbar. 

However, it also breaks the sliding panel transitions, because I had to refrain from rendering panels until their ready to display. Leaving them "off-stage" conflicted with trying to introduce the horizontal scroll bar.

As a follow-up, I'm going to look into things like [ReactCSSTransitionGroup](https://facebook.github.io/react/docs/animation.html#high-level-api-reactcsstransitiongroup) to do trickery like render the panels for a tick and then add a classname to trigger an animation.

Submitting this PR, though, to see if the general shape of the change looks okay

ericawright:
It's a bit difficult to test - I'm trying to figure out how to reproduce the bug, but in the mean time:
The scroll on the manage tabs page (only when it appears on pages that cannot be snoozed) has disappeared.

lmorchard:
Yeah, the bugs only seem to happen like 1/100 attempt to open the panel. I also started [this simpler WebExtension](https://github.com/lmorchard/webextension-toolbar-vs-menu) to isolate the toolbar vs menu issues.

I'll poke at the scrolling on the manage panel some more, though!

ericawright:
I suppose we've decided this is acceptable to cause problems when it is in the menu but, he'se another side effect:
screen shot 2017-03-14 at 1 53 54 pm
going into `manage snoozes` they are cut off and cannot be deleted

lmorchard:
Yeah, it's definitely still very buggy in the menu sidebar. Want to try to at least stabilize it in the toolbar though, because issues like this are worse:

https://discourse.mozilla-community.org/t/snooze-tabs-pane-flickers-on-opening/13872

lmorchard:
Opened this issue to continue working through the problems:

https://github.com/bwinton/SnoozeTabs/issues/263

PR #259, opened 2017-03-13
First merge from l10n branch with locales

### LINKED ISSUE #251: Merge in first batch of new locales
We have Pontoon committing to the l10n branch, and [we've accumulated a bunch of locales](https://github.com/bwinton/SnoozeTabs/compare/l10n?expand=1). We should merge all that work into master and release it!

Also need to tweak experiment page YAML on Test Pilot to reflect the new locales when we're ready.

lmorchard:
Fixes #251

### DIFF
--- a/locales/ar/snoozetabs.properties
+++ b/locales/ar/snoozetabs.properties
@@ -1,0 +1,24 @@
+# This add-on will let you select tabs to close for a while, after which they
+# will re-appear in your browser with a notification.
+
+# Alternative text for the tab icon
+
+# "{time_exp}" is the time the tab will snooze until, see also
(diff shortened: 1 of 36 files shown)


lmorchard:
Has a lot of commits. Trying to think of a better way to do this while retaining all the commit author credits. Squashing might discard that stuff.

lmorchard:
Also as an FYI, [you have change `general.useragent.locale` in `about:config` to switch locales in the WebExtension](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Internationalization#Testing_out_your_WebExtension). Probably have to restart, too, in order to see string changes throughout the add-on

PR #260, opened 2017-03-14
Tweak fatal error handling in pontoon-to-webext script

6a68:
This came up in a Page Shot review that included a copypasta'd pontoon-to-webext.js. Nice idea, thought it seemed worth upstreaming :+1:

### DIFF
--- a/bin/pontoon-to-webext.js
+++ b/bin/pontoon-to-webext.js
@@ -27,6 +27,11 @@ function error(...args) {
   console.error(...args);  // eslint-disable-line no-console
 }
 
+function fatal(...args) {
+  error(...args);
+  process.exit(1);
+}
+
 function getListLocales() {
   return new Promise((resolve, reject) => {
     if (supportedLocales === '*') {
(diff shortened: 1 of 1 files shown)


6a68:
FWIW, Ian also suggested making the `catch` inside `writeFiles` a `fatal` error. That seemed different than a tiny, but nice, refactor, so I've left it out here.

bwinton:
As a side question, should I release pontoon-to-webextension as an npm module of its very own?  I didn't bother, because I figured no-one else would use it, but if it's getting copypasta'd, maybe I can make our lives easier…

lmorchard:
Yeah, if it's turning out to be handy for other WebExtensions, might as well package it up for reuse!

6a68:
@bwinton Yes! It's a really handy tool. I was hoping it'd get published, especially since there's already a not-quite-as-handy tool published under the mozilla org: https://github.com/mozilla/pontoon-to-json

bwinton:
'K.  I'll do that tomorrow morning, before y'all wake up.  :wink:
(In Paris for a DevTools work week, and heading out to dinner soon..)

PR #268, opened 2017-03-15
Re-introduce toolbar vs menu styling

### LINKED ISSUE #263: Panel is not usable when the button is moved from Toolbar to Menu
Opening this as a kind of catch-all issue for problems when moving the button from default Toolbar position into the Menu panel.

lmorchard:
- Move resize width detection into bootstrap script to catch event as
  early as possible

- Only watch for the first resize event, stop listening after (to address [this issue reported on Discourse](https://discourse.mozilla-community.org/t/snooze-tabs-pane-flickers-on-opening/13872))

- More logging from bootstrap script

- Re-add styles removed for #179

- Tweak styles to better use flexbox and absolute sizing on toolbar

Fixes #263.

### DIFF
--- a/.sass-lint.yml
+++ b/.sass-lint.yml
@@ -12,3 +12,4 @@ rules:
   no-vendor-prefixes: 0
   no-qualifying-elements: 0
   nesting-depth: 3
+  no-css-comments: 0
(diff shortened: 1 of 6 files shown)


lmorchard:
So, I know I just made a big stink about moving forward with broken styles in the menu sidebar. But, I spent some time with [webextension-toolbar-vs-menu](https://github.com/lmorchard/webextension-toolbar-vs-menu) yesterday afternoon and I think I managed to work out a few key things that I transferred over to Snooze Tabs. If this PR looks good, I'd like to fit it into the v1.0.15 release rather than ship with the broken styles.

ericawright:
looks good! everything works quite well.

PR #278, opened 2017-03-22
fix: Only build manually selected locales for production

### LINKED ISSUE #275: Several areas are not localized in Asturian [ast]
[Affected versions]:
- Nightly 55.0a1
- Snooze Tabs 1.0.15 (built on 03/20/2017)

[Affected Platforms]:
- All Windows
- All Mac OS
- All Linux

[Prerequisites]:
- Have a Firefox profile with the latest Snooze Tabs add-on version (1.0.15 -dev, built on 03/20/2017) loaded as a temporary add-on.
- Have the Asturian [ast] Firefox locale build installed.

The following areas are displayed in English and should be displayed in Asturian [ast]:
- All the strings in the "Manage SnoozedTabs" panel.
- The "Real Soon Now!", "Next Open" and "Pick a Date/Time" snoozing options.
- "This tab will snooze until" string in the notification bar.
- The "Snooze!" button in the "Pick a Date/Time" panel.
- The "SnoozeTabs" context-menu "Real Soon", "Next Open" options..
- The "SnoozeTabs" tooltip from the "SnoozeTabs" button in the browser toolbar. 

[Notes]:
- Attached a screen shot of the affected areas:
issue_collage

lmorchard:
- New `config.productionLocales` property in `package.json` to
  select locales for production build

- `bin/build-locales.js` helper script to select locales for
  `pontoon-to-webext`

- `cross-spawn` dependency to try to make this work on Windows

Fixes #275.

### DIFF
--- a/bin/build-locales.js
+++ b/bin/build-locales.js
@@ -1,0 +1,23 @@
+#! /usr/bin/env node
+
+const spawn = require('cross-spawn');
+
+let locales = '*';
+if (process.env.NODE_ENV === 'production' &&
+    process.env.npm_package_config_productionLocales) {
+  locales = process.env.npm_package_config_productionLocales;
+}
+
+const result = spawn.sync('pontoon-to-webext', [], {
+  stdio: 'inherit',
+  env: Object.assign(
+    {},
+    process.env,
+    { SUPPORTED_LOCALES: locales }
+  )
+});
+
+if (result.error) {
+  console.error(result.error); // eslint-disable-line no-console
+  process.exit(1);
+}
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "babel-preset-stage-0": "^6.16.0",
     "babel-register": "^6.18.0",
     "chai": "^3.5.0",
+    "cross-spawn": "^5.1.0",
     "eslint": "^3.10.2",
     "eslint-plugin-promise": "^3.4.0",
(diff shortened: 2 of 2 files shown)


bin/build-locales.js
```diff
@@ -0,0 +1,18 @@
...
+}
+
+spawn.sync('pontoon-to-webext', [], {
+  stdio: 'inherit',
+  env: Object.assign(
+    {},
+    process.env,
+    { SUPPORTED_LOCALES: locales }
+  )
+});
```
bwinton: Should we get pontoon-to-webext to just check a key in the npm config (if it exists), so that we could remove this extra file?
…
I guess we would still want to do it only if it was for production, and that doesn't seem like something that pontoon-to-webext would really be interested in, so having this file seems fine.

lmorchard: Yeah, I figured this was more a Snooze Tabs specific hack for npm scripts switching on NODE_ENV==production and using the package.json config. Other projects might want to do this differently (e.g. if they're using gulp or webpack)

lmorchard:
> Assuming spawn.sync will propagate errors from pontoon-to-webext so that the build will stop if we mess something up, I'm good with this change.

Good catch, because it did not! Just tweaked that and updated. Will merge on tests passing

bwinton:
>Will merge on tests passing

:+1:  🙂

PR #284, opened 2017-03-23
fix: Rename previous Snoozed Tabs bookmarks folder on install.

### LINKED ISSUE #237: Update has deleted old snoozed tabs.
Installed the latest snoozetabs via Test Pilot and found that tabs snoozed under previous version issued March 2016 were all deleted & now there is no way to recover them. Warning should be issued to either backup previous tabs prior to installing latest version or move older snoozed tabs to a different folder within bookmarks.

Can I suggest that option be made available to export/back up snoozed tabs to avoid further issues such as this and also to be able to change the default snoozed times/periods.

Otherwise one of my top 3 extensions that is crucial to me web surfing and productivity.

bwinton:
Seems nicer than just deleting it.  We may end up with a lot of Snoozed
Tabs backup folders, but that's probably better than the alternative.

Fixes #237.

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -40,6 +40,20 @@ function updateButtonForTab(tabId, changeInfo) {
 
 function init() {
   log('init()');
+  browser.runtime.onInstalled.addListener((details) => {
+    if (details.reason === 'install') {
+        const title = browser.i18n.getMessage('bookmarkFolderTitle');
+        browser.bookmarks.search({title: title}).then(folders => {
+          return Promise.all(folders.map(folder => {
+            return browser.bookmarks.update(folder.id, {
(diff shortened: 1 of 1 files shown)


src/background.js
```diff
@@ -40,6 +40,20 @@ function updateButtonForTab(tabId, changeInfo) {
...
+    if (details.reason === 'install') {
+        const title = browser.i18n.getMessage('bookmarkFolderTitle');
+        browser.bookmarks.search({title: title}).then(folders => {
+          return Promise.all(folders.map(folder => {
+            return browser.bookmarks.update(folder.id, {
+              title: `${title} - ${getLocalizedDateTime(moment(), 'date_year')} ${getLocalizedDateTime(moment(), 'confirmation_time')}`
+            });
+          }));
+        }).catch(reason => {
+          log('init bookmak folder rename rejected', reason);
```
lmorchard: `s/bookmak/bookmark/` - otherwise :+1:

PR #285, opened 2017-03-23
fix: Get React Storybook working again

### LINKED ISSUE #177: React Storybook stories are broken after l10n work
Since we're using the `browser` API in some interesting places, the React Storybook stories are broken now. I'd like to fix them, because wrangling CSS in the WebExtension is painful

lmorchard:
- Stripped down SnoozePopup component in stories, since that component
  now does a lot of app logic and state management

- Narrow styles less specific to make stories easier to mock

- Mocks for global browser.i18n APIs in `.storybook/config.js`

- Upgrade `@kadira/storybook` dependency

- Typo fix in SnoozePopup component, drop unused `narrowPopup` state

Fixes #177.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "url": "https://github.com/bwinton/SnoozeTabs/issues"
   },
   "devDependencies": {
-    "@kadira/storybook": "^2.21.0",
+    "@kadira/storybook": "2.35.3",
     "@kadira/storybook-addon-knobs": "^1.7.1",
     "babel-cli": "^6.23.0",
     "babel-plugin-external-helpers": "^6.18.0",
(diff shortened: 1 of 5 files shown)


bwinton:
I never got storybook working, so I'm going to pass this one off to my minion…  😉

lmorchard:
Hopefully, you can see this in action with an `npm start` in one console and then `npm run storybook` in another. That should start a server at http://localhost:6006 with previews of the React components

bwinton:
Huh, whaddaya know, it does work for me!

package.json
```diff
@@ -12,7 +12,7 @@
     "url": "https://github.com/bwinton/SnoozeTabs/issues"
   },
   "devDependencies": {
-    "@kadira/storybook": "^2.21.0",
+    "@kadira/storybook": "^2.35.3",
```
bwinton: We just ran into a bunch of problems with inexact versions of npm modules in the Firefox Style Guide, what do you think about pinning this to exactly "2.35.3"?

lmorchard: I can't think of a good reason not to pin it!

PR #290, opened 2017-03-29
fix: Only open a new tab if there are none left after snoozing

### LINKED ISSUE #272: A new empty tab is wrongly opened if only one tab is opened and you change a snoozed tab's date/hour
[Affected versions]:
- Nightly 55.0a1
- Snooze Tabs 1.0.15

[Affected Platforms]:
- All Windows
- All MAC OS
- All Linux

[Prerequisites]:
- Have a Firefox profile with the latest Snooze Tabs add-on version (1.0.15 -dev, built on 17/03/2017) loaded as a temporary add-on.
- Have a snoozed website and an opened tab on the same profile.

[Steps to reproduce]:
1. Open the browser with the profile from prerequisite.
2. Click on the Snooze Tabs button and then on Manage Snoozed Tabs.
4. Click on the date and time.
5. Modify the date/time and click on "Snooze!" button.
6. Observe what happens next.

[Expected result]:
- Manage Snoozed Tabs panel is displayed.

[Actual result]:
- Manage Snoozed Tabs panel is displayed and a new empty tab is opened.

[Notes]:
- Please see the Screen cast attached:
newt

lmorchard:
Fixes #272.

Not sure why this was `<= 1` rather than just `< 1`, but it seems to fix this particular bug. Wondering if I'm missing an edge case here though.

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -159,7 +159,17 @@ const messageOps = {
   },
   update: message => {
     Metrics.updateSnoozedTab(message);
-    return messageOps.cancel(message.old).then(() => messageOps.confirm(message.updated));
+
+    const newId = idForItem(message.updated);
+    const oldId = idForItem(message.old);
+    if (newId === oldId) { return; }
+
+    const toSave = {};
+    toSave[newId] = message.updated;
+    return saveAlarms(toSave)
+      .then(() => removeAlarms(idForItem(message.old)))
+      .then(updateWakeAndBookmarks)
+      .catch(reason => log('confirm rejected', reason));
   },
   setconfirm: message => {
     setDontShow(message.dontShow);


src/background.js
```diff
@@ -133,7 +133,7 @@ const messageOps = {
     delete message.tabId;
     toSave[idForItem(message)] = message;
     return browser.tabs.query({}).then(tabs => {
-      if (tabs.length <= 1) {
+      if (tabs.length < 1) {
```
bwinton: It's `<= 1` because we don't remove the tab until down below, so if there's only one tab, and you snooze it, we won't open an `about:home` tab, and you'll be left with nothing.

lmorchard: Hmm, then maybe this should instead be `if ((tabId && tabs.length <= 1) || tabs.length < 1)` - because we *don't* close any tabs when updating an existing snooze with a new time (which piggybacks on this `confirm` handler). That, or maybe I should add a separate `update` handler rather than re-using `confirm`.

bwinton: I'd be happy with the `if ((tabId && tabs.length <= 1) || tabs.length < 1)` fix.  Heck, we could even make it `if ((tabId && tabs.length === 1) || tabs.length < 1)`, to be extra clear that we're closing the last tab…  Although, that's kinda weird, cause how would we even be clicking on a button in a window with no tabs?

Maybe the real check should be `if (tabId && tabs.length <= 1)`.

lmorchard:
Opted to rework the `update` handler instead, since I realized we were also sending a `cancel` & `confirm` metric event after every `update` event.

src/background.js
```diff
@@ -159,7 +159,14 @@ const messageOps = {
...
     Metrics.updateSnoozedTab(message);
-    return messageOps.cancel(message.old).then(() => messageOps.confirm(message.updated));
+    return removeAlarms(idForItem(message.old))
+      .then(() => {
+        const toSave = {};
+        toSave[idForItem(message.updated)] = message.updated;
+        return saveAlarms(toSave);
+      })
+      .then(updateWakeAndBookmarks)
+      .catch(reason => log('confirm rejected', reason));
```
bwinton: If the remove succeeds, and then the save fails, we'll lose the snoozed tab, which seems suboptimal.
Could we do this in the reverse order instead, so that the failure case will leave two snoozed tabs instead of none?

lmorchard:
Updated to reverse the remove & save. Adds one UI quirk in that both items are there for a split second. Also had to guard against deleting the existing item if the "update" was just re-saving the old without a change.

PR #291, opened 2017-03-30
fix: Add a unique string to the bookmark folder.

### LINKED ISSUE #273: Bookmark management doesn't play well with Sync

bwinton:
To prevent sync fights.

Fixes #273.
Again, sorta.

### DIFF
--- a/locales/en-US/snoozetabs.properties
+++ b/locales/en-US/snoozetabs.properties
@@ -3,7 +3,8 @@
 extDesc=An add-on to let you snooze your tabs for a while.
 popupTitle=Snooze tab until…
 contextMenuTitle=Snooze tab until…
-bookmarkFolderTitle=Snoozed Tabs
+# "{unique_id}"" is an identifier added to the bookmark folder's name to avoid conflicts. Do not translate it.
+uniqueBookmarkFolderTitle=Snoozed Tabs - {unique_id}
 notificationTitle=Tab woke up…
 
 # Alternative text for the tab icon
--- a/src/background.js
+++ b/src/background.js
@@ -42,16 +42,18 @@ function init() {
   log('init()');
   browser.runtime.onInstalled.addListener((details) => {
     if (details.reason === 'install') {
-        const title = browser.i18n.getMessage('bookmarkFolderTitle');
-        browser.bookmarks.search({title: title}).then(folders => {
+      getMetricsUUID().then(clientUUID => {
+        const title = browser.i18n.getMessage('uniqueBookmarkFolderTitle', clientUUID);
+        return browser.bookmarks.search({title: title}).then(folders => {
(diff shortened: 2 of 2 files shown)


bwinton:
Also tagging @rnewman and @flodolo for their thoughts…

locales/en-US/snoozetabs.properties
```diff
@@ -3,7 +3,8 @@
 extDesc=An add-on to let you snooze your tabs for a while.
 popupTitle=Snooze tab until…
 contextMenuTitle=Snooze tab until…
-bookmarkFolderTitle=Snoozed Tabs
+# "{unique_id}" is a per-device id to stop Sync from complaining.
```
flodolo: Localizers are not interested in why you're adding an ID, more if they should localize it or not.

Maybe something like "{unique_id} is an identifier added to the bookmark folder's name to avoid conflicts. Do not translate it"

rnewman:
I think this is fine for the experiment. The ideal is that you get uniqueness in the ID and descriptiveness in the name, but Sync will smush two folders with the same name and parent, so here we are!

locales/en-US/snoozetabs.properties
bwinton: :+1:

src/background.js
```diff
@@ -42,16 +42,18 @@ function init() {
   log('init()');
   browser.runtime.onInstalled.addListener((details) => {
     if (details.reason === 'install') {
-        const title = browser.i18n.getMessage('bookmarkFolderTitle');
-        browser.bookmarks.search({title: title}).then(folders => {
+      getMetricsUUID().then(clientUUID => {
```
lmorchard: I'd say let's not use the metrics UUID for this because it's potentially close to PII, but all I have is vague handwaving and no real scenarios of doom.

PR #292, opened 2017-03-31
fix: Don't wake tabs up in incognito windows.

### LINKED ISSUE #287: Opens on Private Browsing
To avoid awaking my work tabs via the "Restore Session" button when you restart firefox, I right clicked it and selected private browsing. Snooze Tabs reopened one of the pages I had set for today. Perhaps an open for "enable on Private Browsing" would be appropriate, as this feature may also be an advantage, as I may be able to open one tab without restarting my entire session. However, some tabs, especially work related, I do not want to reopen upon opening a separate window

edit: I also cannot seem to put the tab back on snooze for next Tuesday. Being able to 'transfer' tabs from this private browsing (due to the nature in which I use Firefox) would be highly valuable. Sorry, I'm notorious for using programs in unintended and sometimes complicated ways.

This is an excellent management system, though. Very useful for my "schedule."

bwinton:
`Next Open` now opens on the active non-private-browsing window in the next session (or the first non-private-browsing window if the active window is private, or the window it was snoozed from if that window isn't private)…

It's a little complicated to describe, but I think it's doing the right thing in all the cases I could think of…

Fixes #287.

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -15,7 +15,8 @@ import { getLocalizedDateTime } from './lib/time-formats';
 import { NEXT_OPEN, PICK_TIME, times, timeForId } from './lib/times';
 import Metrics from './lib/metrics';
 import { getAlarms, saveAlarms, removeAlarms,
-         getMetricsUUID, getDontShow, setDontShow } from './lib/storage';
+         getMetricsUUID, getDontShow, setDontShow,
+         getNextOpenAlarms } from './lib/storage';
 const WAKE_ALARM_NAME = 'snooze-wake-alarm';
 
 let iconData;
@@ -56,6 +57,7 @@ function init() {
       });
     }
   });
+  browser.windows.onCreated.addListener(handleNextOpen);
   browser.alarms.onAlarm.addListener(handleWake);
   browser.notifications.onClicked.addListener(handleNotificationClick);
   browser.runtime.onMessage.addListener(handleMessage);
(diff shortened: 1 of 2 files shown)


src/background.js
```diff
@@ -93,21 +94,40 @@ function init() {
...
+    if (items.length) {
+      browser.windows.onCreated.addListener(handleNextOpen);
+    }
   }).catch(reason => {
     log('init storage get rejected', reason);
   });
 }
 
+function handleNextOpen(window) {
+  if (!window.incognito) {
```
lmorchard: Could just do `if (window.incognito) { return; }` here and drop a level of indent throughout the rest.

src/background.js
```diff
@@ -93,21 +94,40 @@ function init() {
...
+    getNextOpenAlarms().then(items => {
+      if (items.length) {
```
lmorchard: Same here, an `if (!items.length) { return; }` could de-dent this block

src/background.js
```diff
@@ -93,21 +94,40 @@ function init() {
 
   getMetricsUUID().then(clientUUID => {
     Metrics.init(clientUUID);
-    return getAlarms();
+    return getNextOpenAlarms();
```
lmorchard: Hmm, `getNextOpenAlarms` gets called twice - maybe don't need this first call, since `handleNextOpen` does it again & checks length anyway? Or maybe I'm missing something in the logic here?

src/lib/storage.js
```diff
@@ -24,6 +25,12 @@ export function getAlarms() {
   return getAlarmsAndProperties().then(data => data.alarms);
 }
 
+export function getNextOpenAlarms() {
+  return getAlarms().then(items => {
+    return Object.entries(items).filter(item => item[1].time === NEXT_OPEN);
```
lmorchard: Could use `Object.values` and change `item[1]` to just `item`

src/background.js
bwinton: So, I did it that way so that we wouldn't register the window listener if we didn't need to, but it's probably not that expensive, so I'll remove it.

src/background.js
bwinton: Done.

src/background.js
bwinton: Done.

src/lib/storage.js
bwinton: And done.  🙂

PR #303, opened 2017-04-11
Tweaks to accomodate RTL languages

### LINKED ISSUE #281: On HE build, the panels and confirmation notification are not mirrored in the RTL mode
[Affected versions]:
- Firefox 55.0a1
- Snooze Tabs 1.0.15

[Affected Platforms]:
- All Windows
- All MAC OS
- All Linux

[Prerequisites]:
- Have a Firefox profile with the latest Snooze Tabs add-on version (1.0.15 -dev, built on 22/03/2017) loaded as a temporary add-on.
- Have the Hebrew [he] Firefox locale build installed.

[Steps to reproduce]:
1. Open the localized Firefox build with the profile from prerequisites.
2. Navigate to any website.
3. Click the Snooze Tabs button and observe how the translated items are displayed in the panel.

[Expected result]:
- All items are displayed from right to left (flipped) and correctly translated in Hebrew.

[Actual result]:
- All items are correctly translated in Hebrew, but are wrongly displayed from left to right (normal)

[Notes]:
- This issue also affects the confirmation notification. 
- Please see the Screenshot attached:

rtl

lmorchard:
- Apply dir="{ltr,rtl}" as appropriate to root component in popup

- Add rtl-detect dependency to identify RTL locales

- Style tweaks to properly lay out time options for RTL

- Panel slides in from left for RTL locale

- Tweak SASS lint nesting-depth to work with RTL exceptions

- Fixes for a few propTypes warnings in storybook

Fixes #281.

### DIFF
--- a/.sass-lint.yml
+++ b/.sass-lint.yml
@@ -11,5 +11,5 @@ rules:
   shorthand-values: 0
   no-vendor-prefixes: 0
   no-qualifying-elements: 0
-  nesting-depth: 3
+  nesting-depth: 5
   no-css-comments: 0
--- a/package.json
+++ b/package.json
@@ -134,9 +134,9 @@
     "moment": "2.17.1",
     "rc-calendar": "7.5.1",
     "rc-time-picker": "2.2.1",
+    "react": "15.4.1",
     "react-addons-css-transition-group": "15.4.2",
     "react-dom": "15.4.1",
-    "react": "15.4.1",
     "testpilot-metrics": "2.1.2",
     "uuid": "3.0.1"
   }
--- a/src/lib/utils.js
+++ b/src/lib/utils.js
@@ -5,3 +5,6 @@ export function makeLogger(prefix) {
 }
 
 export const idForItem = item => `${item.time}-${item.url}`;
+
+export const getLangDir = lang =>
+  ['ar', 'fa', 'he'].indexOf(lang) === -1  ? 'ltr' : 'rtl';
--- a/src/lib/components/SnoozePopup.js
+++ b/src/lib/components/SnoozePopup.js
@@ -3,9 +3,10 @@ import ReactCSSTransitionGroup from 'react-addons-css-transition-group';
 
 import MainPanel from './MainPanel';
 import ManagePanel from './ManagePanel';
-import { makeLogger } from '../utils';
+import { makeLogger, getLangDir } from '../utils';
 
 const log = makeLogger('FE <SnoozePopup>');
+const langDir = getLangDir(browser.i18n.getUILanguage());
 
 export default class SnoozePopup extends React.Component {
   constructor(props) {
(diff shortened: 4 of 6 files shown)


lmorchard:
Set `general.useragent.locale` to `he` in `about:config` to test. I *think* this covers things. Seems like there should be more, so I might be missing something.

package.json
```diff
@@ -134,9 +134,10 @@
     "moment": "2.17.1",
     "rc-calendar": "7.5.1",
     "rc-time-picker": "2.2.1",
+    "react": "15.4.1",
     "react-addons-css-transition-group": "15.4.2",
     "react-dom": "15.4.1",
-    "react": "15.4.1",
+    "rtl-detect": "1.0.0",
```
bwinton: That's a _lot_ of code to pull in for something that should be a simple check against an array, or css selector `:-moz-locale-dir(rtl)`…

bwinton:
(Even if we have to set the `dir` attribute, I'd prefer we set it on the body, and then checked for `body[dir='rtl'] div.whatever`, instead of setting it on a bunch of sub-elements.)

lmorchard:
The `dir` attribute only gets set on the root element of the SnoozePopup component

package.json
lmorchard: [It's actually a tiny module](https://github.com/shadiabuhilal/rtl-detect/blob/master/lib/rtl-detect.js). I was just going to do a simple array lookup, but this module seemed to do some nice coverage of language & country in locale codes. I *could* probably just run our list of locales through it and capture the result as a hardcoded lookup hash

lmorchard:
Updated to get rid of rtl-detect. Seeing if I can move the dir attrib to body and still work with the styles in storybook

bwinton:
Oh, I was confused by having to set it on the `div` _and_ the `ReactCSSTransitionGroup` for both the `SnoozePopup.js` and the `stories/index.js`…  Perhaps it would be okay if it was on the SnoozePopup if it's hard to move it to the body.

.sass-lint.yml
```diff
@@ -11,5 +11,5 @@ rules:
   shorthand-values: 0
   no-vendor-prefixes: 0
   no-qualifying-elements: 0
-  nesting-depth: 3
+  nesting-depth: 5
```
bwinton: Do we still need the nesting depth change?  5 seems like a lot…

lmorchard: Yeah, there's 5 levels of depth on the new styles. Some interaction with other rules forces a construct like `div { &[dir=rtl] }` and then another level for the narrow menu case like `.narrow { div { &dir[div=rtl] { } } }`

bwinton: Okay, cool.

src/lib/utils.js
```diff
@@ -5,3 +5,6 @@ export function makeLogger(prefix) {
 }
 
 export const idForItem = item => `${item.time}-${item.url}`;
+
+export const getLangDir = lang =>
```
lmorchard: One last tweak here to further simplify the language direction thing, since I had this code copypasta'd in two places

lmorchard: Planning to hit the green merge button if CircleCI likes everything

PR #310, opened 2017-04-18
L10n for 1.0.17

lmorchard:
Can't directly merge l10n into master and push, so here's a cut from l10n branch for the v1.0.17 release.


lmorchard:
Kind of wish I had a cleaner way to squash this into one commit, but I don't want to risk losing the author credits. Hmm

PR #319, opened 2017-04-22
Update snoozetabs.properties

WhiteSymmetry:


lmorchard:
@flodolo Just checking: Does Pontoon work okay with merging a pull request like this?

bwinton:
I think we might want this sort of thing to go through Pontoon, to let the localizers confirm it says what we think it says…

flodolo:
@WhiteSymmetry
Thanks for the contribution and for taking the time to file a request, but I would like this kind of changes to go through the team and the tool they're using (also cc @selimsumlu)
https://wiki.mozilla.org/L10n:Teams:tr

Tool: https://pontoon.mozilla.org/tr/

On a side note, this is not the branch where localization lives.

PR #333, opened 2017-07-06
Fetch 100% locales from pontoon instead of hardcoding

meandavejustice:


lmorchard:
We actually don't want to do this - we want to manually enable new locales when they're 100% and ready. IIRC, some 100% locales are still being worked on despite all strings being localized

meandavejustice:
@lmorchard There are "suggested" and "translated" strings, `completed` should only grab 100% translated strings. We can double check with flod tho.

lmorchard:
Yeah, I think it was flod who advised not to automate this... trying to find the issue we closed on it

lmorchard:
Going to close this one for now, because we want to manually select production locales.

lmorchard:
FWIW, also just added a pile of new locales in https://github.com/bwinton/SnoozeTabs/pull/387

PR #336, opened 2017-07-24
Chore: Bump to v1.0.19

lmorchard:


lmorchard:
Let's get #337 merged before this, so I can push out the l10n updates in 1.0.18

lmorchard:
We can merge this to bump the version for another deployment, then I'll submit another PR for a second version bump to catch any work we do next

PR #338, opened 2017-07-24
fix: Rework add-on build & packaging for new Mozilla Extension signing

lmorchard:
- `npm run package` now produces `addon.xpi` per new Mozilla Extension
  signing conventions

- Update `bin/build-addon.sh` to expect `addon.xpi` on dev builds

- Leave the signing code alone in `bin/build-addon.sh`, which *should* continue to work for "legacy" signing against AMO

- `circle.yml` moves `*.xpi` rather than `*.zip` into artifacts


lmorchard:
Putting out the batsignal for reviews - I think this *should* make Snooze Tabs work with the new signing while still working with the old signing.

PR #349, opened 2017-08-07
Ensure production build on `npm run package`

lmorchard:
- Also add `npm run package:dev` for development builds


lmorchard:
Looks like we accidentally got a dev build in the switch to the new signing infra. This should fix it.

package.json
```diff
@@ -114,7 +115,8 @@
     "rollup:snooze-bootstrap": "rollup -c --environment entry:popup/snooze-bootstrap",
     "rollup:confirm-bar": "rollup -c --environment entry:lib/confirm-bar",
     "bundle:css": "npm run lint:sass && shx mkdir -p dist/popup && node-sass src/popup/snooze.scss > dist/popup/snooze.css",
-    "package": "npm run build && web-ext build -s dist && mv web-ext-artifacts/*.zip addon.xpi",
+    "package": "cross-env NODE_ENV=production npm run build && web-ext build -s dist && mv web-ext-artifacts/*.zip addon.xpi",
+    "package:dev": "cross-env NODE_ENV=development npm run build && web-ext build -s dist && mv web-ext-artifacts/*.zip addon.xpi",
```
bwinton: Should we move this to `addon-dev.xpi` instead, to help avoid sending the wrong xpi?

lmorchard: Oh, good idea! Also updated the build-addon.sh script still used in CircleCI to use these NPM scripts. Also still used by the old signing infra, but I think that's going to fall away soon (but not yet?)

PR #352, opened 2017-08-09
fix: Change name from "snoozetabs" to "Snooze Tabs"

### LINKED ISSUE #342: Snooze Tabs add-on name displayed like a single word and with lowercase in "about:addons" page
**[Affected versions]:**
- Firefox 54.0.1 and above

**[Affected Platforms]:**
- All Windows
- All Mac
- All Linux

**[Prerequisites]:**
- The "testpilot.env" string type pref was manually added in "about:config" page with "dev" value.
- The "extensions.webapi.testing" boolean type pref was manually added in "about:config" page with "true" value.
- Test Pilot and all the active experiments were installed.

**[Steps to reproduce]:**
1. Open the Firefox with the profile from prerequisite and navigate to "about:addons" page.
2. Click on "Extensions" tab from left side.
3. Observe the names of all the Test Pilot related add-ons.

**[Expected result]:**
- All the experiments have the name displayed with uppercase letters at the beginning of each word

**[Actual results]:**
- Snooze Tabs add-on has the name displayed as "snoozetabs".

**[Notes]:**
- Here is a screenshot of the issue:
snoozetab name

lmorchard:
Fixes #342.

### DIFF
--- a/bin/generate-manifest.js
+++ b/bin/generate-manifest.js
@@ -4,17 +4,16 @@ const path = require('path');
 const fs = require('fs');
 const packageMeta = require('../package.json');
 
-const manifest = Object.assign(packageMeta.webextensionManifest, {
-  'name': packageMeta.name,
+const manifest = Object.assign({
   'author': packageMeta.author,
   'version': packageMeta.version,
   'homepage_url': packageMeta.homepage,
   'applications': {
-    'gecko': Object.assign(packageMeta.webextensionManifest.applications.gecko, {
+    'gecko': Object.assign({
       'id': packageMeta.id
-    })
+    }, packageMeta.webextensionManifest.applications.gecko)
   }
-});
+}, packageMeta.webextensionManifest);
 
 const outPath = path.join(path.dirname(__dirname), 'dist', 'manifest.json');
 fs.writeFileSync(outPath, JSON.stringify(manifest, null, '  '));
--- a/package.json
+++ b/package.json
@@ -75,6 +75,7 @@
   },
   "webextensionManifest": {
     "manifest_version": 2,
+    "name": "Snooze Tabs",
     "applications": {
       "gecko": {
         "update_url": "https://testpilot.firefox.com/files/snoozetabs@mozilla.com/updates.json"


package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "snoozetabs",
+  "name": "Snooze Tabs",
```
pdehaan: I don't think package.json files can have spaces in the `name` property.
I think we may have to set the `title` property: https://developer.mozilla.org/en-US/Add-ons/SDK/Tools/package_json

lmorchard: Hmm, that's a page on the defunct SDK that doesn't apply to webextensions. In this case, I think it's fine because npm's package.json rules only say leading & trailing spaces are invalid.

lmorchard:
Rebased to get current against master

bin/generate-manifest.js
```diff
@@ -14,7 +14,7 @@ const manifest = Object.assign(packageMeta.webextensionManifest, {
       'id': packageMeta.id
     })
   }
-});
+}, packageMeta.webextensionManifest);
```
bwinton: So, this will make it so that the name in `webextensionManifest` will override the name we specify here, which makes me wonder why we are still specifying the name up on line 8…

bin/generate-manifest.js
```diff
@@ -14,7 +14,7 @@ const manifest = Object.assign(packageMeta.webextensionManifest, {
       'id': packageMeta.id
```
bwinton: Should we also change this to have the `packageMeta.webextensionManifest.applications.gecko` come after the `'id'`, to be consistent with line 17?

package.json
```diff
@@ -74,6 +74,7 @@
   },
   "webextensionManifest": {
     "manifest_version": 2,
+    "name": "Snooze Tabs",
```
bwinton: Should/can this be localized, too, now that we're not using the one from the outer scope?

lmorchard: I think we've been avoiding localizing product / project names in general

bin/generate-manifest.js
lmorchard: Hmm, looks like I was kind of overthinking it in case this script gets used for another add-on and the name *doesn't* need to be overridden. But, npm has all kinds of restrictions on name that are inconvenient for an add-on name that we might as well always specify it

bin/generate-manifest.js
lmorchard: Hmm, also here I think I thought having the package.json ID override the webextensionManifest data was a good idea. But I can't remember why - seems like a bad idea right now

lmorchard:
Made some tweaks, but left out localizing the add-on title

lmorchard:
🎉

PR #360, opened 2017-08-18
fix: Constrain confirm bar CSS to element ID

### LINKED ISSUE #340: Clicking one of the snoozing options badly breaks website's UI
[Notes]:
- This is a regression that is present in Nightly and Beta.
- Here is the regression range:
Last good revision: cee08328318a1ba8adcc1eff9c783fbecfed3540
First bad revision: 659451f14dddb66b1bf2c27fc19b704a3fd4e64a
- Pushlog:
https://hg.mozilla.org/integration/autoland/pushloghtml?fromchange=cee08328318a1ba8adcc1eff9c783fbecfed3540&tochange=659451f14dddb66b1bf2c27fc19b704a3fd4e64a

[Affected versions]:
- Beta 55.0b11
- Nightly 56.0a1

[Affected Platforms]:
- All Windows
- All Linux
- All Mac

[Prerequisites]:
- Have latest Snooze Tabs v. 1.0.18 experiment installed.

[Steps to reproduce]:
1. Open the browser with the profile from prerequisites and navigate to www.youtube.com
2. Click on the Snooze Tabs button from toolbar.
3. Click on one of the snoozing options and observe the effect on the website's UI.

[Expected result]:
- Website's UI elements remain unchanged.

[Actual result]:
- Website's UI breaks.

[Additional notes]:
- This seems to affect most of the tested websites (Google, Reddit, Yahoo, Facebook, Wikipedia etc)
- Added a screen recording with the issue:

snooze break

lmorchard:
Looks like scoped CSS is not working as expected, so let's brute-force
this to select just for the confirm bar CSS.

Fixes #340.

### DIFF
--- a/src/lib/confirm-bar.js
+++ b/src/lib/confirm-bar.js
@@ -41,8 +41,8 @@ chrome.runtime.onMessage.addListener(function({message, iconData, closeData}) {
     document.body.appendChild(confirmationBar);
   }
   confirmationBar.outerHTML = `<div id="${confirmationId}" dir="${langDir}">
-    <style scoped>
-      * {
+    <style>
+      #${confirmationId} * {
         margin: 8px 0 8px 8px;
       }
       #${confirmationId} {
@@ -67,11 +67,11 @@ chrome.runtime.onMessage.addListener(function({message, iconData, closeData}) {
       #${confirmationId}.shown {
         top: 0;
       }
-      img {
+      #${confirmationId} img {
         height: 24px;
         width: 24px;
       }
-      button {
+      #${confirmationId} button {
         padding: 0;
         background-color: #fbfbfb;
         border: 1px solid #b1b1b1;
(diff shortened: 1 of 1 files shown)


src/lib/confirm-bar.js
```diff
@@ -42,7 +42,7 @@ chrome.runtime.onMessage.addListener(function({message, iconData, closeData}) {
   }
   confirmationBar.outerHTML = `<div id="${confirmationId}" dir="${langDir}">
     <style scoped>
```
bwinton: Should we also remove the `scoped`, or leave it in as a fallback?

src/lib/confirm-bar.js
```diff
@@ -67,11 +67,11 @@ chrome.runtime.onMessage.addListener(function({message, iconData, closeData}) {
       #${confirmationId}.shown {
         top: 0;
       }
-      img {
+      #${confirmationId}.img {
```
bwinton: Should that `.` be a ` `?  (Uh, that second bit is a space character…)

src/lib/confirm-bar.js
lmorchard: Hmm, yeah, seeing in a few places ([like MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/style#Attributes)) that scoped was deprecated awhile ago and is now disappearing from browsers 😞

PR #384, opened 2018-01-11
fix snooze tabs to be more photony

### LINKED ISSUE #368: Refresh look and feel of Snooze Tabs to better match Photon
A few visual tweaks to the Snooze Tabs UI based on the Photon Design System will allow it to better fit in with the look & feel of Firefox going forwards.

https://mozilla.invisionapp.com/share/EAEBBAEG2#/262298173_Artboard

### LINKED ISSUE #381: Add dark theme icon

johngruen:
Fixes #368 
Fixes #381


johngruen:
@lmorchard circle ci is failing here for reasons that don't have to do with my PR, any ideas?

package.json
```diff
@@ -30,6 +30,7 @@
     "node-sass": "3.13.0",
     "npm-run-all": "3.1.2",
     "onchange": "3.2.0",
+    "photon-colors": "^1.3.1",
```
bwinton: We prefer to use strict versions instead of caret ranges,

src/icons/later_today.svg
```diff
@@ -3,7 +3,7 @@
 <svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
 	 viewBox="0 0 24 24" style="enable-background:new 0 0 24 24;" xml:space="preserve">
 <style type="text/css">
-	.st0{fill:#53C942;}
+	.st0{fill:#30e60b;}
```
bwinton: I guess we can't use the `photon-colors` variables here, but could we at least add a comment indicating which one it is, to make it easier to upgrade?

bwinton: Also, we seem to have standardized on uppercase hex colors.

src/popup/snooze.scss
```diff
@@ -1,29 +1,32 @@
 @import 'node_modules/rc-calendar/assets/index';
 @import 'node_modules/rc-time-picker/assets/index';
+@import 'node_modules/photon-colors/colors.scss';
 
 $panel-width: 320px;
 $panel-shadow-width: 13px;
+$white: #fff;
```
bwinton: We don't have `white` and `black` in the photon-colors?!?

src/popup/snooze.scss
```diff
@@ -1,29 +1,32 @@
...
+$black: #000;
+$transparent-black: rgba(0, 0, 0, .05);
```
bwinton: To be more like photon-colors, how about changing this to `$black-a50`?

src/popup/snooze.scss
```diff
@@ -32,31 +35,33 @@ body {
 }
 
 #app {
+  height: 392px;
   width: 320px;
-  height: 424px;
+  background: #e4e4e4;
```
bwinton: Could we use `$grey-20` or `$grey-30` here instead?

lmorchard:
> circle ci is failing here for reasons that don't have to do with my PR, any ideas?

Looking into it a bit this morning, Snooze Tabs has bit-rotted somewhat since last release

src/icons/bell_icon.svg
```diff
@@ -4,7 +4,7 @@
 	 x="0px" y="0px" width="24px" height="24px" viewBox="0 0 24 24"
 	 style="enable-background:new 0 0 24 24;" xml:space="preserve">
 <style type="text/css">
-	.st0{fill:#595959;}
+	.st0{fill:context-fill}
```
lmorchard: Is this option going to work if/when we go to AMO? I thought this was reserved for mozilla-signed addons?

Also, have a different approach over here in #388 which isn't necessarily mutually exclusive to this one

johngruen:
Finally getting back to this.

src/popup/snooze.scss
johngruen: apparently not :)

src/icons/bell_icon.svg
johngruen: @lmorchard here's the DXR for this. Do we have to change add-on IDs b/c that's how the permission is set

https://dxr.mozilla.org/mozilla-central/source/layout/svg/SVGContextPaint.cpp#69

lmorchard: Okay, so it relies on the add-on ID and not the signing key. So, from there I guess my question is: Should we still be leaning on a special case that other WebExtensions don't have if/when this thing leaves Test Pilot?

lmorchard:
Let's just go ahead with the context-fill thing (versus my theme detection stuff) since it's simpler for now

lmorchard:
Hmm, "dismissing" a review seems way ruder than I meant it 😟

bwinton:
🤣

PR #388, opened 2018-01-12
fix: Switch browser action icon based on light / dark theme

### LINKED ISSUE #67: In DevEdition's dark theme, the SnoozeTabs icon should be white

lmorchard:
As of Fx56, we have access to the [`browser.management` API](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/management/getAll). That lets us enumerate themes and find the enabled one. Doesn't give us light vs dark, but we can try to make an educated guess from the IDs.

- Light / dark bell icons for enabled / disabled states

- Attempt to use the "theme_icons" property in manifest.json for an
  initial icon while background.js inits

- Start a lists of light / dark themes, since themes themselves don't
  identify as light or dark

- Detect theme & update browser action button icon on tab switch &
  update, as well as theme change

Fixes #67.


SoftVision-PaulOiegas:
@lmorchard If I'm not mistaking, default buttons color on dark theme is `BEBEBE` while this fix makes Snooze Tabs button color to be `B1B1B1`. It's close but not quite there.

lmorchard:
Updated the icon color. I'd taken a screenshot and used the eyedropper in Gimp to get the color, but I must have been too close to the edge of an icon when I did it and got an anti-alias color rather than the main color

SoftVision-PaulOiegas:
@lmorchard Bell color for Firefox Dark theme is a match now after your fix. 👍 However, and I hope you won't curse me too much. Last night when I checked this, I didn't checked also the Default and Light themes. It looks like we have a difference on those too...
- Default theme buttons color `4A4A4B` vs implemented `595959`
- Light theme buttons color `5A5B5C` vs implemented `595959`.

Sorry again for not checking this.

lmorchard:
For those light / default theme icons, I kind of settled on a near match compromise, since I don't think we want to have different icons exactly matching every different theme. Main thing was that it wasn't visible on the dark theme

johngruen:
@lmorchard would moving to amo mean we'd lose the @mozilla.org signature?

lmorchard:
> would moving to amo mean we'd lose the @mozilla.org signature?
'
Yes, we only get signed by the special Mozilla key when using the Test Pilot build infrastructure. We lose that if/when we move to AMO

lmorchard:
Let's just go ahead with #384 and context-fill for simplicity for now

Issue #322, opened 2017-04-28
Snoozed tabs didn't open


scheinercc:
I had two tabs snoozed until today, one for 5pm, one for 6pm. Around 6:30pm I realised that they haven't opened yet. I opened one via "Manage snoozed tabs", navigated to a new page and only when I saved this tab for a later date the other two opened simultaneously. First time that happened. 
System: 55.0a1 (2017-04-27) (64-bit) on macOS

lmorchard:
Hmm, we've had a similar report before (#234) but have not been able to reliably reproduce the issue to find the cause. I'll leave this open to see if anyone else sees it again. You don't happen to have any browser console error messages, do you? (Probably not, since I'd be surprised if you had it open at the time in normal use)

As for the sudden opening of the late snoozed tabs - we've got code that opens up everything that's due or overdue. That happens when a) a timer goes off or b) anything changes the set of snoozed tabs (e.g. when snoozing, updating, or deleting). 

Sounds like at some point you lost the timer that should have gone off for 5pm & 6pm, but I don't know why that would have happened.

albinhh:
I had similar strange behaviour as well. I didn't make much of it, because I thought it was due to "user error" but I will try to reproduce it next time it happens and post my findings.

scheinercc:
Will keep an eye out and report it here if it happens again, too.

jibe-b:
I have the same behaviour since 2 or 3 days.

lmorchard:
> I have the same behaviour since 2 or 3 days.

Did you happen to see any error messages in the log?

jibe-b:
May I ask which log I should check?

lmorchard:
> May I ask which log I should check?

The browser console - under the Tools > Web Developer > Browser Console menu

latk:
I may have managed to reproduce this or a related issue:

 1. Snooze a tab for n minutes (n >= 2) by manually selecting a time.
 2. Suspend the browser, e.g. by hibernating the host OS. Issuing a SIGSTOP to Firefox should also work.
 3. Wait for > n minutes. The timer should have fired during this time, but obviously can't since the process is not running.
 4. Resume Firefox. Expectation: The snoozed tab appears within a few seconds.
 5. Wait for n minutes.
 6. The snoozed tab appears.

I checked this with n = 2 minutes, under FF 55.0.2, using Ubuntu 16.04 as host OS.

[...]

PR #390, opened 2018-01-12
Add a periodic alarm every minute as a backup for wake timer

lmorchard:
Not sure if this will absolutely fix alarm reliablilty reports we've
gotten. But, operating on the notion that maybe we're losing the timer
for the next tab that needs to be woken, this commit adds a timer that
fires every minute and causes any pending tabs to wake.

Issue #322.


lmorchard:
This is just kind of a desperate handwave to make sure tabs wake up. The wake alarm handler checks for pending items and does nothing if none are due to wake, so it shouldn't hurt to call it more often than ideally necessary.

src/background.js
```diff
@@ -269,9 +276,9 @@ function handleWindowCreated(window) {
   }
 }
 
-function handleWake() {
+function handleWake(alarm) {
   const now = Date.now();
-  log('woke at', now);
+  log('woke at', now, 'with alarm', alarm ? alarm.name : 'none');
```
bwinton: Could we add some telemetry to see if this catches any alarms that wouldn't have fired?
(I think it's a good idea to add, but would like to know how necessary it is…)

src/background.js
```diff
@@ -260,6 +262,11 @@ function updateWakeAndBookmarks() {
     });
 }
 
+function startPeriodicAlarm() {
+  log('starting periodic alarm');
+  return browser.alarms.create(PERIODIC_ALARM_NAME, { periodInMinutes: 1 });
```
bwinton: Will the `return browser.alarms.clearAll()` at line 241/243 clear this alarm, too? And if so, do we need to re-create it in that function?

lmorchard: Oof yeah, I think that's right. Will look at that

src/background.js
lmorchard: Hmm, maybe can think of adding an event for late alarms? Will have to refresh my metrics-fu to figure out how to do taht

src/background.js
lmorchard: Pushed a tweak to just specifically clear WAKE_ALARM_NAME and not the periodic alarm.

lmorchard:
Going to merge this and submit a follow-up issue to think about adding metrics to measure alarm lateness on wake.

PR #392, opened 2018-01-12
fix: Stop using async script loader in popup

lmorchard:
WebExtension popup panel performance has improved to the point that this
hack to mitigate slow-down is now apparently causing a slow-down. This PR
removes the hack and allows the popup to load the JS directly.


lmorchard:
> Do you have any numbers on how much removing this speeds it up?

Not really, I've never been able to figure out how to measure performance there. Subjectively speaking, though, before this patch I could see the "..." placeholders in time stamps of snooze choices. After the patch, it looks like things render almost immediately

PR #395, opened 2018-01-17
fix: Move confirmation bar into an iframe to further isolate styles

### LINKED ISSUE #340: Clicking one of the snoozing options badly breaks website's UI

### LINKED ISSUE #371: Various issues regarding the confirmation bar on Send website
[Affected versions]:
- Firefox 56.0.2 and above

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- Have latest Snooze Tabs v. 1.0.21 experiment installed.
- "Ask for confirmation when snoozing tabs" checkbox is checked.

[Steps to reproduce]:
1. Open the browser with the profile from prerequisites and navigate to https://send.firefox.com/
2. Click on the "Snooze tabs" toolbar button.
3. Click on one of the snooze options (except "Pick a Date/Time").
4. Observe the location of the confirmation bar.
5. Observe the UI and the elements of the confirmation bar.
6. Click on "Cancel" button and observe what happens next. 

[Expected result]:
4. The confirmation bar is placed on the top of the page content.
5. An orange SnoozeTabs icon aligned to the left followed by the text "This tab will be snooze until [snoozed option]". "OK" and "Cancel" buttons are also displayed and at the right margin "Don't show this again" text along with an unchecked checkbox and an "X" close button.
6. The confirmation bar is dismissed.

[Actual results]:
4. The confirmation bar is displayed on the left bottom part of the page content.
5. The UI is broken and the orange SnoozeTabs icon and "X" button are not displayed.
6. Nothing happens, the confirmation bar remains displayed.

[Notes]:
- The issue is also reproducible when installing snooze-tabs from Test Pilot.
- Attached a screencast of the issue:

snooze notif bar

### LINKED ISSUE #380: Confirmation UI is broken on Deadspin
screen shot 2018-01-10 at 7 49 14 pm

### LINKED ISSUE #383: Consider moving confirmation into the panel
B/C the injected content is fragile, we should maybe just do the confirmation dialog in the panel. @sevaan wanna to a mock for this?

lmorchard:
Mostly stole the idea for this from [the notification bar in bitwarden](https://github.com/bitwarden/browser/blob/master/src/content/notificationBar.js#L340), but hacked it a bit to preserve our dynamic HTML for l10n. Seems to work and should more effectively isolate styles

Fixes #383.
Fixes #380.
Fixes #371.
Fixes #340.

### DIFF
--- a/src/lib/confirm-bar.js
+++ b/src/lib/confirm-bar.js
@@ -21,6 +21,13 @@ function sendShowUpdate(showCheckbox) {
   }
 }
 
+function removeConfirmBar() {
+  const el1 = document.getElementById('snoozetabs-confirm-bar');
+  if (el1) { el1.remove(); }
+  const el2 = document.getElementById('snoozetabs-confirm-bar-spacer');
+  if (el2) { el2.remove(); }
+}
+
 chrome.runtime.onMessage.addListener(function({message, iconData, closeData}) {
   const atTime = confirmationTime(message.time, message.timeType);
   const confirmationId = 'snoozetabs-confirmation-bar';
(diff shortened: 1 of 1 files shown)


src/lib/confirm-bar.js
```diff
@@ -35,120 +42,145 @@ chrome.runtime.onMessage.addListener(function({message, iconData, closeData}) {
...
-  const okButton = document.getElementById(okId);
-  okButton.addEventListener('click', () => {
-    browser.runtime.sendMessage({
-      'op': 'confirm',
-      'message': message
+  removeConfirmBar();
+
+  const iframe = document.createElement('iframe');
+  iframe.style.cssText = 'height: 45px; width: 100%; border: 0;';
+  iframe.id = 'snoozetabs-confirm-bar-iframe';
```
bwinton: So, if we're putting it in an iframe, could we set the `.src` attribute to load a page from our extension, so that we can remove the inline html from this file?

src/lib/confirm-bar.js
```diff
@@ -35,120 +42,145 @@ chrome.runtime.onMessage.addListener(function({message, iconData, closeData}) {
...
+      });
+      confirmationBar.classList.toggle('shown');
+      removeConfirmBar();
+      sendShowUpdate(showCheckbox);
+    });
+
+    const cancelButton = iframeDocument.getElementById(cancelId);
+    cancelButton.addEventListener('click', () => {
+      confirmationBar.classList.toggle('shown');
+      window.setTimeout(removeConfirmBar, 300);
```
bwinton: Why do we need to remove the confirm bar in this code, and not in the previous code?

src/lib/confirm-bar.js
```diff
@@ -35,120 +42,145 @@ chrome.runtime.onMessage.addListener(function({message, iconData, closeData}) {
...
+    }, 500);
+
+    const okButton = iframeDocument.getElementById(okId);
+    okButton.addEventListener('click', () => {
+      browser.runtime.sendMessage({
+        'op': 'confirm',
+        'message': message
+      });
+      confirmationBar.classList.toggle('shown');
+      removeConfirmBar();
```
bwinton: Should this also be a `window.setTimeout(removeConfirmBar, 300)`, like in cancel and close?
For that matter, do we need to do this at all, since the tab will be going away if they click the Ok button? 😉

src/lib/confirm-bar.js
```diff
@@ -35,120 +42,145 @@ chrome.runtime.onMessage.addListener(function({message, iconData, closeData}) {
...
+      confirmationBar.classList.add('shown');
+    }, 500);
+
+    const okButton = iframeDocument.getElementById(okId);
+    okButton.addEventListener('click', () => {
+      browser.runtime.sendMessage({
+        'op': 'confirm',
+        'message': message
+      });
+      confirmationBar.classList.toggle('shown');
```
bwinton: Should these be `toggle` or `remove`? I guess it's the same, since you won't be able to click them if it's not shown, although if you click cancel then ok really quickly, then maybe you could get them both to fire… Anyways, we use `add` above, so maybe these should all be `remove` for consistency.

src/lib/confirm-bar.js
lmorchard: We could, but I didn't because of all the strings that need to be injected from browser.i18n.getMessage. Alternative would be to load static HTML with all blank elements, then set the text content for every element needing it.

src/lib/confirm-bar.js
lmorchard: They were originally toggle, so I didn't change them. Could be remove instead

src/lib/confirm-bar.js
lmorchard: Yeah, probably doesn't need to be here because the tab closes. That's actually why I didn't put a timer on the removal, which only make half sense in retrospect

src/lib/confirm-bar.js
lmorchard: You mean previous as in before the iframe? I think we should have been doing it there, too. The markup hangs around in the DOM that we probably should have cleaned up, even though things were hidden by styles

src/lib/confirm-bar.js
bwinton: Okay, let's leave it as you have it then.

src/lib/confirm-bar.js
```diff
@@ -35,120 +42,143 @@ chrome.runtime.onMessage.addListener(function({message, iconData, closeData}) {
...
+      confirmationBar.classList.remove('shown');
+      window.setTimeout(removeConfirmBar, 300);
+      sendShowUpdate(showCheckbox);
+    });
+
+    const closeButton = iframeDocument.getElementById(closeId);
+    closeButton.addEventListener('click', () => {
+      confirmationBar.classList.remove('shown');
+      window.setTimeout(removeConfirmBar, 300);
+      sendShowUpdate(showCheckbox);
```
bwinton: So this function looks a lot like the one for the cancel button. They're both kinda short, so I don't know if it makes sense to merge them, or leave it duplicated, but if that's something you wanted to do, I would be okay with it. :)

lmorchard: Yeah, that's worth a quick DRYing up too

PR #398, opened 2018-01-18
fix: Remove confirm bar spacer, which does not seem needed

lmorchard:


lmorchard:
Well, I stole it from the Bitwarden iframe code, and it does push down content. But, it doesn't seem to be necessary since the confirm bar is position: fixed and visible in any case. The *site* content does get covered, though. Going go either way on this PR, though, and close it rather than merge

PR #404, opened 2018-01-25
Fix ID in webext manifest

### LINKED ISSUE #401: The Snooze Tabs icon is wrongly colored in black
[Affected versions]:
- Release 58.0 and up
- Snooze Tabs 1.0.21

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- Have a Firefox profile with the latest Snooze Tabs add-on version (https://665-15314951-gh.circle-artifacts.com/0/tmp/circle-artifacts.PS80CkV/addon.xpi) loaded as a temporary add-on.
- Default theme is enabled.

[Steps to reproduce]:
1. Open the browser with the profile from prerequisite and navigate to a website.
2. Observe the "Snooze Tabs" icon from the toolbar.

[Expected result]:
- The icon is grey.

[Actual result]:
- The icon is black.

[Notes]:
- This issue is also reproducible with the icons in the context menu and in "about:addons" page.
- Attached a screenshot of the issue:

black icon

lmorchard:
Looks like the ID for the WebExtension in the manifest.json was left
off, which means the extension was getting a generated ID.

That ID didn't end in @mozilla.com, which means that the [context-fill
feature][context-fill] feature wasn't enabled.

This also would have caused other problems.

[context-fill]: https://dxr.mozilla.org/mozilla-central/source/layout/svg/SVGContextPaint.cpp#69

Fixes #401.

### DIFF
--- a/bin/generate-manifest.js
+++ b/bin/generate-manifest.js
@@ -8,12 +8,9 @@ const manifest = Object.assign({
   'author': packageMeta.author,
   'version': packageMeta.version,
   'homepage_url': packageMeta.homepage,
-  'applications': {
-    'gecko': Object.assign({
-      'id': packageMeta.id
-    }, packageMeta.webextensionManifest.applications.gecko)
-  }
 }, packageMeta.webextensionManifest);
+
+manifest.applications.gecko.id = packageMeta.id;
 
 const outPath = path.join(path.dirname(__dirname), 'dist', 'manifest.json');
 fs.writeFileSync(outPath, JSON.stringify(manifest, null, '  '));


bin/generate-manifest.js
```diff
@@ -8,12 +8,9 @@ const manifest = Object.assign({
   'author': packageMeta.author,
   'version': packageMeta.version,
   'homepage_url': packageMeta.homepage,
-  'applications': {
-    'gecko': Object.assign({
-      'id': packageMeta.id
-    }, packageMeta.webextensionManifest.applications.gecko)
-  }
 }, packageMeta.webextensionManifest);
```
lmorchard: Turns out that [the change I made earlier that changed the order of merging these objects](https://github.com/bwinton/SnoozeTabs/commit/4c4bbfd68d3aa1e924ec220e6a8e137b282a4f40#diff-ae48bfc07bcae523cc9f374db73f6a1cL7) just ended up clobbering the `applications.gecko.id` assignment altogether.

bwinton: Could we remove the id from `webExtensionManifest.applications.gecko` instead, if we know we want the packageMeta one?

lmorchard: [Problem is that there isn't an id in there already](https://github.com/bwinton/SnoozeTabs/blob/master/package.json#L80). The only property in there is `update_url`, and that was clobbering the merged object

bwinton: Oh, cause they don't get merged… Dang. Okay, r=me.

lmorchard: Yeah, `Object.assign` is a dumb flat merge ugh

PR #415, opened 2018-01-29
Quick fixes for new QA issues

lmorchard:
This covers a handful of new issues, though I could probably use some help with the CSS changes. Particularly around the colors


src/popup/snooze.scss
```diff
@@ -652,7 +652,7 @@ li.entry {
     border-radius: 2px;
     font-size: 14px;
     text-align: center;
-    width: 6em;
+    width: 7.85em;
```
lmorchard: This seems pretty hacky... My CSS-fu is failing me in getting the widths on the time selector right.

src/icons/confirm_bell_icon.svg
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Generator: Adobe Illustrator 19.0.0, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
+<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"
+	 x="0px" y="0px" width="24px" height="24px" viewBox="0 0 24 24"
+	 style="enable-background:new 0 0 24 24;" xml:space="preserve">
+<style type="text/css">
+	.st0{fill:#ff1ad9;}
```
lmorchard: Couldn't find a pink version of the bell, so I eye-dropper sampled this pink from the image in the updates photon spec.

src/lib/confirm-bar.js
```diff
@@ -91,23 +92,23 @@ chrome.runtime.onMessage.addListener(function({message, iconData, closeData}) {
...
           font-family: Lucida Grande, Tahoma, sans-serif;
-          font-size: 12px;
-          height: 24px;
+          font-size: 13px;
+          height: 32px;
           width: 6em;
         }
         #${confirmationId} button:hover {
-          background-color: #ebebeb;
+          background-color: #f9f9fa;
```
lmorchard: These hover & active colors for the cancel button on the confirm bar don't seem right, but the existing ones just kind of merged into the background

bwinton: @sevaan ?

lmorchard:
I just tweaked the color a bit to go from 0.1 to 0.15 opacity using the original color, which seems like a way less dramatic change. Think I'll go ahead and merge this, and we can tweak it again as needed

sevaan:
The Design System provides this info: https://design.firefox.com/photon/components/buttons.html#default-4

image

lmorchard:
Haven't actually seen those design system docs before, so that's handy thanks!

PR #423, opened 2018-02-10
feat: Add option for next weekday when it is a weekend

### LINKED ISSUE #393: Turn "This Weekend" into "Next Weekday" on a weekend
Postponing tabs until the weekend is great, but likewise when it is the weekend I have tabs that I want to postpone until Monday (e.g. work things). I can use the Custom Date option, but given "This Weekend" doesn't make sense when it is already the weekend, maybe it could turn into "Next Weekday".

kewisch:
This fixes #393, might as well DIY :)

I've added an icon from the noun project and increased the line width ever so slightly to make it match better with the other icons. I'm not a designer so it might still be off, if someone else wants to give it a shot please do!


src/lib/times.js
```diff
@@ -41,6 +42,10 @@ export function timeForId(time, id) {
       rv = rv.day(6).hour(9).minute(0);
       text = getLocalizedDateTime(rv, 'short_date_time');
       break;
+    case 'weekday':
+      rv = rv.isoWeekday(8).hour(9).minute(0);
```
bwinton: We're already using `.day` everywhere, so let's keep consistent with that, please. (Also up in the `hideif` functions.)

sevaan:
I put my comments in #393, but will include them here:

"I'm not sure this is necessary. Next Week always points to the upcoming Monday, and "Next Weekday" would just do the same thing. I think a better fix would be to make This Weekend say "Next Weekend", point to the upcoming Saturday and include the date in the row."

src/lib/times.js
kewisch: I used `isoWeekday` since it goes from 1 = Monday to 7 = Sunday. `day` goes from 0 = Sunday to 6 = Saturday. So to get the next Monday, I would need to differentiate and use `.day(1)` if it is Sunday, and `.day(8)` on any other day.

Similar in `hideif`, where I can just use `>` to get the weekend instead of comparing to `0` or `6`

PR #426, opened 2018-02-15
a11y and glitch fixes

johngruen:
This is mildly WIP, but resolves both of the visual glitches as well as most of the A11y issues in the ESR branch. I will file separate (non-blocking) issues for remaining a11y concerns.

### DIFF
--- a/src/lib/components/ManagePanel.js
+++ b/src/lib/components/ManagePanel.js
@@ -25,7 +25,6 @@ export default class ManagePanel extends React.Component {
   render() {
     const { id, entries, active, tabIsSnoozable, dontShow, updateDontShow, moment } = this.props;
     const { undoEntries, datepickerActive } = this.state;
-
     const sortedEntries = Object.values({...entries, ...undoEntries});
     sortedEntries.sort((a, b) => {
       if (a.time === NEXT_OPEN) {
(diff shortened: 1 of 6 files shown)


src/lib/components/ManagePanel.js
```diff
@@ -25,7 +25,7 @@ export default class ManagePanel extends React.Component {
   render() {
     const { id, entries, active, tabIsSnoozable, dontShow, updateDontShow, moment } = this.props;
     const { undoEntries, datepickerActive } = this.state;
-
+    console.log(entries); //eslint-disable-line no-console
```
pdehaan: Did we want to leave this in, or ✂️ ?

src/lib/components/ManagePanel.js
```diff
@@ -155,19 +172,35 @@ export default class ManagePanel extends React.Component {
     return !active || datepickerActive;
   }
 
-  handleBack() {
+  shouldFocus(undo = false) {
+    return (undo || this.shouldIgnoreClicks()) ? -1 : 1;
+  }
+
+  shouldIgnoreKeyEvents(ev) {
+    console.log(ev.key); //eslint-disable-line no-console
```
pdehaan: ✂️ ?

pdehaan:
Also, not sure if it's worth adding the [**eslint-plugin-jsx-a11y**](https://www.npmjs.com/package/eslint-plugin-jsx-a11y) ESLint plugin and extending the "plugin:jsx-a11y/recommended" ruleset and see if anything interesting pops up.

lmorchard:
> Not sure if you want to commit the package-lock.json file or not. I don't see it currently in the repo, but I also don't see it in .gitignore. ¯\_(ツ)_/¯

Not part of the repo, but didn't exist as a thing when this project started - which is why it's not in .gitignore. We could add it, but not too worried about it.

lmorchard:
Looks pretty good, but yeah: needs the console.log statements removed. [There's a logger util over here](https://github.com/bwinton/SnoozeTabs/blob/master/src/popup/snooze-content.js#L13) if you really want the messages in there - it shuts itself off on a production build

Issue #32, opened 2016-12-12
Keyboard navigation


lmorchard:
Ensure keyboard navigation in the popup is supported

sevaan:
From Issue #63:

> Keyboard navigation possible improvements (probably needs a separate issue). Initial focus is correct - the whole panel is focused. Consequent focus should go to the list of possible snooze options. And after that to the bottom menu bar. Next TAB should blur the panel completely and go to the next widget in the browser. I would suggest the following interactions:
> 
>- Current active element should always be visible to the user.
>- When the panel has a list of options or a list of snoozed tabs user should navigate the list with up and down arrows and the focus should be maintained on the list container. Pressing enter should do the same things as clicking on the list item. (when on the manage panel, this will help with automatically showing delete button when list item is selected with a keyboard). Delete should have a keyboard shortcut when delete button is pressed.
>- Calendar widget should maintain keyboard focus on grid container and internal navigation (within the grid) should happen via arrow keys.

lmorchard:
Kicking out of a milestone for now

Issue #488, opened 2021-08-25
Can't click snooze time in Firefox 93.0a1


gaul:
I cannot select "Later today", "Tomorrow", etc. using Firefox nightly 93.0a1 (2021-08-24).  This is a recent regression.

bwinton:
Well, the code didn't change on my end, so I suspect it's a WebExtension bug. Can you see if there's anything in the Browser Console, and file it [on Bugzilla](https://bugzilla.mozilla.org/enter_bug.cgi?product=WebExtensions&component=General) and post the link here so that I can follow it?

gaul:
Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1727958.

gaul:
This is fixed upstream.

PR #490, opened 2021-08-28
Add keyboard shortcuts to sleep panel

gaul:
This allows the up and down arrows to select a sleep duration and
enter to confirm it.  It also works around a focus regression in
Firefox Nightly 93 and matches the Multi-Account Containers behavior.
References #488.  References #32.


src/popup/snooze.html
```diff
@@ -6,7 +6,21 @@
...
+      <div class="panel-wrapper">
```
bwinton, line 9: I wonder why it was all on one line previously… Perhaps something about extra whitespace messing with the styling? Could you add a comment with screenshots from before and after this patch is applied?

src/popup/snooze-content.js
```diff
@@ -126,13 +126,44 @@ const resizeHandler = () => {
...
+  switch (e.keyCode) {
+  case 40: {
+    const nextElement = selectables[index + 1];
+    if (nextElement) {
+      nextElement.focus();
+    }
+    break;
+  }
+  case 38: {
+    const previousElement = selectables[index - 1];
```
bwinton, line 143: Perhaps, if index is already negative, we should move to the bottom element, and have them wrap? What do you think?

bwinton (changes requested):
Seems mostly good, but there are a couple of things I'd like to get fixed before I merge it…

bwinton:
I also just noticed that the up/down arrows will continue to do stuff when the calendar widget is open, which they shouldn't. Maybe we should do the processing in `lib/components/ManagePanel.js` instead, to make sure it's scoped correctly…

bwinton:
(Also, this kinda fixes #32, which is nice! 😃)

src/popup/snooze-content.js
```diff
@@ -126,13 +126,44 @@ const resizeHandler = () => {
   });
 };
 
+const keyboardHandler = (e) => {
+  const selectables = [...document.querySelectorAll('.option')];
+  const element = document.activeElement;
+  const index = selectables.indexOf(element) || 0;
+
+  switch (e.keyCode) {
+  case 40: {
```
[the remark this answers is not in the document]
gaul: Done.

src/popup/snooze-content.js
```diff
@@ -126,13 +126,44 @@ const resizeHandler = () => {
...
+    break;
+  }
+  case 38: {
+    const previousElement = selectables[index - 1];
+    if (previousElement) {
+      previousElement.focus();
+    }
+    break;
+  }
+  case 39:
```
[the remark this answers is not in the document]
gaul: Done.

src/popup/snooze-content.js
gaul: Done.

src/popup/snooze.html
gaul: Removed unneeded change.

gaul:
> I also just noticed that the up/down arrows will continue to do stuff when the calendar widget is open, which they shouldn't. Maybe we should do the processing in `lib/components/ManagePanel.js` instead, to make sure it's scoped correctly…

Could you give me some more guidance on this?  I don't understand where I would add the keyboard handler.

bwinton:
> > I also just noticed that the up/down arrows will continue to do stuff when the calendar widget is open, which they shouldn't. Maybe we should do the processing in `lib/components/ManagePanel.js` instead, to make sure it's scoped correctly…
> 
> Could you give me some more guidance on this? I don't understand where I would add the keyboard handler.

Yeah, so on [this line](https://github.com/bwinton/SnoozeTabs/blob/main/src/lib/components/ManagePanel.js#L70), there's an `onKeyPress` handler, which forwards all the keys to [this function](https://github.com/bwinton/SnoozeTabs/blob/main/src/lib/components/ManagePanel.js#L182-L188). So maybe we can add an `onKeyPress` to the div on [this line](https://github.com/bwinton/SnoozeTabs/blob/main/src/lib/components/ManagePanel.js#L45), and handle the keys there? (We'd need to allow them to pass through [this function](https://github.com/bwinton/SnoozeTabs/blob/main/src/lib/components/ManagePanel.js#L178-L180), but that shouldn't be so bad…

gaul:
> Yeah, so on [this line](https://github.com/bwinton/SnoozeTabs/blob/main/src/lib/components/ManagePanel.js#L70), there's an `onKeyPress` handler, which forwards all the keys to [this function](https://github.com/bwinton/SnoozeTabs/blob/main/src/lib/components/ManagePanel.js#L182-L188). So maybe we can add an `onKeyPress` to the div on [this line](https://github.com/bwinton/SnoozeTabs/blob/main/src/lib/components/ManagePanel.js#L45), and handle the keys there? (We'd need to allow them to pass through [this function](https://github.com/bwinton/SnoozeTabs/blob/main/src/lib/components/ManagePanel.js#L178-L180), but that shouldn't be so bad…

Sorry, I spent a few hours on this and probably won't figure it out myself.  I suspect you mean `MainPanel.js` though?

bwinton (approved):
Well, let's not let the perfect be the enemy of the good.
👍🏼 from me, and we can move the code around later. 🙂

PR #493, opened 2021-08-29
Handle reader-mode tabs, by storing an extra bit of info in the message.

### LINKED ISSUE #424: Can't snooze a tab in read mode
When a tab is in read mode, the snooze icon is grayed and I can't snooze it. 
I have to leave the read mode.

bwinton:
Fixes #424.
@rugk, @gaul: Either of you want to take a look at this and give it a 👍🏼/👎🏼?

### DIFF
--- a/src/popup/snooze-content.js
+++ b/src/popup/snooze-content.js
@@ -10,7 +10,7 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import { getAlarmsAndProperties } from '../lib/storage';
 import SnoozePopup from '../lib/components/SnoozePopup';
-import { makeLogger } from '../lib/utils';
+import { makeLogger, getUrl } from '../lib/utils';
 
 const log = makeLogger('FE');
 log('loaded');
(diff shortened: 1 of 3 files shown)


rugk (approved):
Code-wise this LGTM, but what does thus achieve?

Could not you (previously) just re-open tabs in reader mode as they were… in reader mode again. Or does not this work?

**Edit:** Ah forget it, you do that, there is a `openInReaderMode` flag, awesome! That looks nice. :+1:

src/popup/snooze-content.js
```diff
@@ -53,9 +54,11 @@ function undeleteSnoozedTab(item) {
 }
 
 function openSnoozedTab(item) {
+  log('openSnoozedTab', item);
```
rugk, line 57: Is this log intentional or remaining from debugging?

rugk:
BTW watched PRs now, so you don't need to ping me everytime… :wink:

src/popup/snooze-content.js
bwinton: Whoops! Removed!

PR #509, opened 2022-11-30
Use a more intuitive label for "next open"

waldyrious:
I find the "Next open" label quite confusing. The fact that this entry requires a comment suggests that others do too. Here I'm proposing an alternative label, "Next Browser Launch", which I believe may be more intuitive.

I'm marking this PR as a draft because if the change is accepted, I would add additional changes, including renaming the actual variable in the code and removing the comment in the translation files.


bwinton:
That seems reasonable. Can I get a screenshot of how it looks?

waldyrious:
> Can I get a screenshot of how it looks?

I could produce one, but I would have to build this locally and test it. Not sure when I'd be able to look into that.

By the way, does installing the extension build locally interfere with the actual extension in my browser?

bwinton:
Yes, but you can run `npm run build && npm run run` to fire it up in a new profile for testing. 🙂

waldyrious:
Static screenshot:
Screenshot from 2022-12-01 08-04-40

Screen recording:
Screen recording 2021-12-01 08-03-57

bwinton:
Does it also change it in the panel the drops down when you click the bell? (That's the piece I'm more worried about, because of the size constraints when we put that button into the overflow menu…)

waldyrious:
Hmm, maybe I'm looking in the wrong place but the panel doesn't seem to be tight for space at all:

Ag11i5VG7q

bwinton:
Oh! Nice… It's been a while since I ran it there.
In that case, I approve this change, so feel free to do the renaming and comment removing!   🙂

waldyrious:
@bwinton ok, I've made the changes. Please take a look when you get the chance.

Note: NPM automatically removed empty `requires` fields in `package.lock`. Since that doesn't seem to affect the actual behavior of the project, I included those changes in a separate commit.

bwinton (approved):
Looks good. Thank you!

Issue #528, opened 2023-10-31
High Memory usage in `about:memory`


Ryman:
I'm pretty sure I'm debugging this poorly, but according to `about:memory` this extension's background page seems to have almost 3GB of memory allocated for icons or images? (going by [this page](moz-extension://97faa3dd-ef6d-4c5e-adb4-65f4aae2d154/) referencing this addon) I have around 65 snoozed tabs currently.

![image](https://github.com/bwinton/SnoozeTabs/assets/994978/34e2311d-dd14-43f0-8048-8b335091bb63)

Do you have any insight into how I could help investigate this further, or am I reading this page wrong?

bwinton:
I don't know how you could investigate it further, but I suspect that all the animated gifs probably take up a bunch of memory…Still there should ideally be only 9 of them at any given time, right?

jacktose:
@Ryman I barely understand this, but I'm concerned about reducing Fx memory usage. Is this what you're looking at? about:memory, extension section, filtered for extension guid:
<img width="1662" height="1077" alt="screenshot of about:memory, extension section, filtered for `d17b3bbd-39ba-4427-8a78-51c8c7f1d89c`" src="https://github.com/user-attachments/assets/92469ae0-c086-4451-b173-e152ea196505" />
That's only totaling 12 MB right now. Am I looking at the wrong thing, or just not snoozing tabs with huge images?

Ryman:
> about:memory, extension section, filtered for extension guid

Yep, it should show up there. I imagine it's a mixture of uptime and what kind of favicon's you're encountering. My session's been running for ~3 hours (which aligns with around 180 copies, each from a sync per minute logging if my theory on https://github.com/bwinton/SnoozeTabs/pull/530 is correct):

<img width="1187" height="775" alt="Image" src="https://github.com/user-attachments/assets/c4bd2498-90f3-429a-881b-134d46a52b5e" />

And if I expand the "tiny" section:

<img width="1148" height="1037" alt="Image" src="https://github.com/user-attachments/assets/e51d1cba-e155-46cb-8b5f-6318f719ae0d" />

Ryman:
One thing worth pointing out, when I was debugging this with my build of the extension in a fresh instance, I sometimes saw these allocations in the main process allocation instead of extensions process (but remnants in both). I can only assume that's related to me using the development console in some way?

PR #530, opened 2025-10-24
fix(memory): reduce memory leaks from logging large objects frequently

Ryman:
Helps alleviate https://github.com/bwinton/SnoozeTabs/issues/528. Logging these objects don't seem to get collected by GC even when using the minimize memory option in `about:memory`. The icons logged with each item can be several kilobytes each and some sites have jpeg or png which can be very large and don't have any benefit to logging. In my case I've over 300 snoozed tabs, which adds up to gigabytes over a few hours.

I've also adjusted some of the arguments in `package.json` as I couldn't start the project using node v16.8 (and the project didn't build due to dependency resolution even with a lockfile on v22.12).

### Reproduction 
For reproduction on the old code run the sync log in a loop and wait a few minutes as the sync happens every minute.

```js
for (let i = 0; i < 20000; i++) { log('Sync Folder!' + i, snoozeTabsFolder, Object.values(items)); }
```

You should be able to filter `about:memory` using the "Memory report path filter" input with the prefix of one of the icons e.g. `svg+xml;base64,PHN` for the Hacker News favicon. Example below (stat totals are based on filtered content):

image


bwinton (approved):

olantwin:
@Ryman With your permission, I'd like to get this merged. As the maintainer I can modify your branch, but I'd still like to ask your permission before making use of that power.

Some parts of your PR are now no longer needed (e.g. the fixes to the tests), but the main change is very valuable. I think there are also a few other large objects we want to avoid logging to reduce memory.

Please let me know!

PR #534, opened 2026-07-20
Build the confirm bar even when the iframe load event does not fire

### LINKED ISSUE #533: stopped working
Stopped working on Debian Firefox 48.0.2 (64-bit). The options are available, but the tab is not snoozed, either when I use the right click menu or the add-on preferences. A pity because it has been a great add-on

olantwin:
Fixes #533

Since Firefox ~149/150, picking a snooze time does nothing — no error anywhere, the tab just stays open. The add-on still appeared to work on addons.mozilla.org / support.mozilla.org.
It does still work, if the "ask for confirmation" checkbox is not set.

This hinted at the underlying cause: `confirm-bar.js` builds the confirmation bar inside an `iframe.onload` handler on a dynamically created blank iframe. In recent Firefox the iframe's initial `about:blank` document is already complete when the iframe is inserted, so the `load` event never fires and the bar is never built. Because the actual snooze (`op: 'confirm'`) is only sent when OK is clicked in that bar, snoozing silently did nothing. On sites where content-script injection is forbidden (AMO, SUMO), `tabs.executeScript` rejects and `background.js` falls back to confirming directly — which is why snoozing kept working exactly there.

Given those findings, this PR attempts a fix by keeping the `load` listener for older Firefox (140 ESR should still use the old behaviour), but also build the bar immediately when `iframe.contentDocument.readyState` is already `complete`, with a guard so it can't be built twice.

I tested this on Firefox 152.0.4 (Linux (NixOS)) with bundles built with the repo's webpack config, loaded via `web-ext run` — the bar appears again and OK snoozes the tab.

Since I'm new to Firefox addons, I'd appreciate feedback and I'm more than happy to iterate on the fix in case a different fix would be better or improvements could be made.

Thanks for a great addon, which at least for me, has become absolutely vital to my Firefox workflow.

### DIFF
--- a/src/lib/confirm-bar.js
+++ b/src/lib/confirm-bar.js
@@ -52,8 +52,9 @@ chrome.runtime.onMessage.addListener(function({message, confirmIconData, closeDa
   frameDiv.appendChild(iframe);
   document.body.appendChild(frameDiv);
 
-  iframe.onload = () => {
+  const buildConfirmBar = () => {
     const iframeDocument = iframe.contentDocument;
+    if (iframeDocument.getElementById(confirmationId)) { return; }
     let confirmationBar = iframeDocument.createElement('div');
     confirmationBar.id = 'snoozetabs-confirmation-bar';
     iframeDocument.body.appendChild(confirmationBar);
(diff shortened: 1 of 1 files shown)


olantwin:
@bwinton May I ask you for a review on this? Then the functionality of this add-on should be restored for those who have not disabled the confirmation on snoozing.

A note on timing, in October: 140 is EOL, so then this PR simplifies even further as only one branch is needed.

bwinton (approved):

bwinton:
Looks good!
So, the biggest problem now is that I'm not set up to do a release of this anymore, and haven't used it myself in a few years… How do you feel about taking over maintainership, so that you can make sure it continues to work for you?

olantwin:
Thanks a lot for merging this!

With the caveat that I mostly come from scientific computing and this is my first contribution to a Firefox addon; this is an addon that is pretty close to my heart and small in scope, so I'd be happy to learn everything necessary to maintain it and keep it working for anyone still using it (and try to tackle/triage the open issues and PRs).

Starting out I might have to ask you some stupid questions from time to time for things I can't understand from AMO documentation or figure out myself, but I'll try to not bother you more than absolutely necessary.

PR #536, opened 2026-08-27
Replace scss with css

olantwin:
Two steps for easy review (most of the diff lines come from the lockfile, O(100)):

1. Move SCSS to CSS to make sure git detects the move.
2. Do the fairly mechanical edits, I did not find a feature used in the SCSS which isn't supported in plain CSS.

To go with this migration, sass-lint is replaced with stylelint.

In a quick local check with Firefox 153 everything looks exactly the same as before.

It appears that now the blocker for a more recent node version is webpack 4 (again, probably a cascade of others after that, but that's to be expected).


stories/index.js
```diff
@@ -5,9 +5,6 @@ import { host } from 'storybook-host';
 import moment from 'moment';
 import { getLangDir } from '../src/lib/utils';
 
-// TODO: Get sass working with storybook
```
bwinton, line 8: Hah! Awesome!

package.json
```diff
@@ -114,16 +114,16 @@
     "copy:assets": "shx cp -r src/icons dist/ && svgo dist/icons --quiet && shx cp -r src/popup/*.html src/popup/FiraSans-Regular.* dist/popup/",
     "bundle:manifest": "node ./bin/generate-manifest",
     "bundle:js": "npm run lint:js && webpack",
-    "bundle:css": "npm run lint:sass && shx mkdir -p dist/popup && node-sass src/popup/snooze.scss > dist/popup/snooze.css",
+    "bundle:css": "npm run lint:css && shx mkdir -p dist/popup && shx cat node_modules/photon-colors/colors.css node_modules/rc-calendar/assets/index.css node_modules/rc-time-picker/assets/index.css src/popup/snooze.css > dist/popup/snooze.css",
```
bwinton: It would be nice if there was an include syntax, but this seems fine for now.

bwinton (approved):
Looks good to me. Let's land it once CI passes…

package.json
olantwin: Looking again, it seems that this also now might be possible in plain CSS. I'll update this PR tomorrow.

olantwin: This is now handled via

https://github.com/bwinton/SnoozeTabs/blob/ef92739ad30dcb94fbb864aff6d85571517528b8/src/popup/snooze.html#L5-L7 , which makes the concatenation unnecessary.

olantwin:
@bwinton If you have time to review the new commits, I'd appreciate it. Only a few lines changed overall.

I'd squash commits [53cd328](https://github.com/bwinton/SnoozeTabs/pull/536/commits/53cd32848282721334af6b71b646c6a6c136f42f)..[3ce924e](https://github.com/bwinton/SnoozeTabs/pull/536/commits/3ce924e45275d02e9df628f5a6457a75a208080d) before merge to have a bisectable history.

README.md
```diff
@@ -6,6 +6,11 @@
 
 An add-on to let you snooze your tabs for a while.
 
+## Requirements
+
+* Node.js 16.x
+* Firefox 117 or later
```
bwinton, line 12: Do we not work in Chrome, or is this dev requirements?

bwinton (approved):
One question, non-blocking. Thanks for the work making the changes!

README.md
olantwin: I guess in theory this should also work on Chrome, but I've not yet checked this is practice. There is another "snooze tabs" addon though (proprietary), so me might need to rename for listing it on the Chrome Addon store.

Once the new release is out of the door I'll have a look.

olantwin: #539

PR #537, opened 2026-08-27
Remove the dead Google Analytics reporting.

olantwin:
metrics.js beaconed every snooze, wake, resnooze, cancel, tab focus, tab close and popup open to ssl.google-analytics.com. Universal Analytics has been shut down, so those requests went nowhere. The tracking ID also belongs to a Mozilla property this fork does not control, so the add-on was sending user telemetry, with no opt-in, to somebody else's account.

Drop the module, its tests, its call sites in the background page, and the two popup messages whose handlers did nothing but report. Removing Metrics.init() also drops its tabs.onActivated and tabs.onRemoved listeners, which only fed the wake history state machine, and the eager getMetricsUUID() call at startup: the UUID is now created lazily on the first bookmark sync instead.

metrics.js held the only .json import under src/, so the json-loader webpack rule and devDependency go too. That import had been inlining the whole of package.json, dependency list and all, into the background bundle.

The data collection disclosure becomes obsolete.

storage.js is deliberately untouched. getMetricsUUID() is not only a telemetry client ID: the same UUID names the per-install bookmark sync folder, so changing its value would orphan every existing user's folder. The key keeps its misleading name until it can be renamed with a robust migration, alongside the move to the new add-on ID.


package.json
```diff
@@ -70,7 +69,6 @@
     "url": "git+https://github.com/bwinton/SnoozeTabs.git"
   },
   "config": {
-    "GA_TRACKING_ID": "UA-90911916-2",
```
bwinton, line 73: 💖

bwinton (approved):
LGTM!

olantwin:
Thanks, I appreciate you having a look!

PR #541, opened 2026-08-31
Fix several CSS bugs discovered during migration

olantwin:
A few bugs/correctness issues discovered while porting from SASS to plain CSS.

Apart from the transition, none should have a visual effect, but I might as well fix them.


src/popup/snooze.css
```diff
@@ -57,7 +52,7 @@ body {
   opacity: .3;
   position: absolute;
   top: 0;
-  transition: left 250ms ease-in-out, opacity 250ms ease-in-out, visibility 250 ease-in-out;
+  transition: left 250ms ease-in-out, opacity 250ms ease-in-out, visibility 250ms ease-in-out;
```
bwinton, line 55: Is there a `transition: start` so that it works for rtl languages, too?

bwinton (approved):

src/popup/snooze.css
```diff
@@ -699,7 +694,7 @@ li.entry {
 div {
   &[dir='rtl'] {
     .panel {
-      left: calc(0 - 100%);
+      left: -100%;
       transition: left 250ms ease-in-out, opacity 250ms ease-in-out;
```
olantwin, line 698: @bwinton I think this line is the one that matters for RTL, which overrides the LTR line above if I'm not mistaken (and misses the `visibility`, which is another thing worth fixing).

olantwin:
@bwinton I think I found some more drift between RTL and LTR. Most of which can be avoided by moving to logical properties (e.g. `inset-inline-start`) to reduce the duplication needed to the minimum between the two cases. I'll draft a separate PR for that and track it in https://github.com/bwinton/SnoozeTabs/issues/542 for now.

PR #543, opened 2026-09-02
Express directional styles with logical properties

### LINKED ISSUE #542: Use logical CSS properties to simplify RTL support.
There is drift between the many properties that currently are defined twice. Using `inset-inline-start` etc.. should cut down on what we need to duplicate (and keep in sync!)

olantwin:
RTL was handled by a hand-mirrored div[dir='rtl'] block that restated every directional rule with the sides swapped. Inline-axis properties let one declaration cover both directions, collapsing the block to the #pick arrow's rotation, the only rule with no logical equivalent I could find.

LTR rendering is unchanged. RTL gains the three rules which had gotten out of sync/were never translated.

Potential follow-up: The calendar arrows weren't changed. Not sure whether these are conventionally fixed for RTL, feedback welcome!

Implements #542.

### DIFF
--- a/src/popup/snooze.css
+++ b/src/popup/snooze.css
@@ -46,34 +46,34 @@ body {
   display: flex;
   flex-direction: column;
   height: 100%;
+  inset-inline-start: 100%;
   justify-content: space-between;
-  left: 100%;
   margin: 0;
   opacity: .3;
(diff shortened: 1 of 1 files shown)


bwinton (approved):
Nice!

PR #544, opened 2026-09-02
Modernise the build stack

olantwin:
The build ran on no Node newer than 16 because webpack 4 hashes with md4 through the OpenSSL provider, which OpenSSL 3 refuses outright with ERR_OSSL_EVP_UNSUPPORTED. This restricted us to a node release that went end-of-life in 2023. babel-preset-env was configured with no targets, so everything compiled to ES5, which prevented us from using async calls.

Version changes:

  babel 6 -> @babel/* 7 (8 blocked by old React)
  babel-loader 7 -> 10
  eslint 4 -> 9 (10 blocked by elint-plugin-react)
  node 16 -> 24
  mocha 10 -> 11
  webpack 4 -> 5
  webpack-cli 3 -> 6
  .babelrc -> babel.config.js
  .eslintrc.json -> flat eslint.config.js

babel.config.js is now the single source of truth for both consumers and sets an explicit target: Firefox for the bundle, current node for the mocha run. Keep that target in step with strict_min_version.

stage-0 is no longer needed, nor is .npmrc, legacy-peer-deps no longer set.

We now specify the oldest supported Firefox version (140 ESR as of
writing) as target.

No behaviour change expected or observed in local testing.


package.json
```diff
@@ -12,24 +12,25 @@
     "url": "https://github.com/bwinton/SnoozeTabs/issues"
   },
   "engines": {
-    "node": ">=16.0.0 <17.0.0"
+    "node": ">=20.0.0"
```
bwinton: Just as a side note, I'm a fan of pinning dependencies to a specific version, since we don't really know what's going to change (or break) between them, and they should probably get some testing before we update. (Also, so that we build and release the same thing we test locally.)

webpack.config.js
```diff
@@ -4,7 +4,7 @@ const webpack = require('webpack');
 const NODE_ENV = process.env.NODE_ENV || 'development';
 
 module.exports = {
-  mode: NODE_ENV,
+  mode: NODE_ENV === 'production' ? 'production' : 'development',
```
bwinton: Would this check be better done up on line 4?

bwinton (approved):
👍🏼

package.json
olantwin: Good point, especially since our plugin is a leaf package that nothing else depends on (for non-leaf packages I remember long discussions in python whether or not to use upper bounds...)

I'll check what versions we can use while also supporting the oldest supported Firefox version.

webpack.config.js
olantwin: Actually, this is just duplicating the check in line 4, if we assume only `NODE_ENV` can only be `production` or `development`. I missed that check, so I'd just remove this one here.

bwinton: I think keeping the check might be nice, since someone could accidentally set `NODE_ENV` to `jfdjslkfjdlsfl` because a cat walked on their keyboard or something… 😉

olantwin: Reading up a bit more, I also saw that NODE_ENV could be unset, so having the fallback to `development` is probably worth having. I'll combine line 4 with the fallback from line 7.

olantwin:
Note: This does not update all deps yet, react and others will need updating in a second step. Trying to break things into manageable chunks.

olantwin:
Everything pinned now to explicit versions and the check for NODE_ENV has been updated..

PR #545, opened 2026-09-03
chore(deps): add dependabot config

olantwin:
Check for new versions weekly, default cooldown of three days. Update PRs will need manual testing before merge.


bwinton (approved):
👍🏼

